PR #8, opened 2018-07-05
Various updates and fixes

### LINKED ISSUE #7: ARC toolchain support
Since the following commit https://github.com/crosstool-ng/crosstool-ng/commit/62f5b90e2e79a8ccee07534f9484627e5e268d50 it's possible to build multilibbed Elf32 (AKA baremetal) toolchain for ARC.

Latest Binutils (2.30), GCC (8.1.0) and GDB (8.1) are recommended to be used for ARC.
But given versions mentioned above are all available in up-to-date Crosstool-NG and moreover they are default versions now there should be no problem with it.

nashif:
- Add ARC (Fixes #7)
- Build IAMCU correctly
- Use latest version of xtools

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+bin/*
+share/*
+build/*
--- a/.shippable.yml
+++ b/.shippable.yml
@@ -12,6 +12,7 @@ env:
         - TARGET=arm
         - TARGET=riscv32
         - TARGET=mips
+        - TARGET=arc
 
 build:
     cache: true
--- a/.shippable.yml
+++ b/.shippable.yml
@@ -21,13 +21,13 @@ build:
         - ${SHIPPABLE_BUILD_DIR}/sources
     pre_ci_boot:
         image_name: zephyrprojectrtos/ci
-        image_tag: v0.3
+        image_tag: v0.4-rc7
         pull: true
         options: "-e HOME=/home/buildslave --privileged=true --tty --net=bridge --user buildslave"
 
     ci:
       - ccache -s
-      - sudo apt-get install -y gawk wget git-core diffstat unzip texinfo gcc-multilib build-essential
+      - sudo apt-get install -y gawk wget git-core diffstat unzip texinfo gcc-multilib build-essential libtool-bin
       - sudo apt-get install -y libreadline-dev gperf flex bison libncurses5-dev texinfo help2man
       - unset CC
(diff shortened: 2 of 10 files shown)


abrodkin:
@nashif I guess you were able to build Zephyr OS for EMSK board and built image works as expected? The reason to ask is our port of newlib in upstream is not very complete yet and if Zephyr doesn't implement its own libc there might be a good reason to use our fork of [newlib](https://github.com/foss-for-synopsys-dwc-arc-processors/newlib) until all of our fixes are accepted upstream.

nashif:
> @nashif I guess you were able to build Zephyr OS for EMSK board and built image works as expected? The reason to ask is our port of newlib in upstream is not very complete yet and if Zephyr doesn't implement its own libc there might be a good reason to use our fork of newlib until all of our fixes are accepted upstream.

yes, the toolchain works fine to the most part and based on the initial testing I have done, one thing that needs fixing in zephyr is the linker script, we need to add `_SDATA` section as per https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/wiki/Small-Data

I will look into the newlib fork above.

abrodkin:
@nashif How did you do with OE-based tools before in regard to `_SDATA` symbol in the linker script?
As for newlib - do you really use it in Zephyr at all? Otherwise no need to spend time on our newlib fork.
That's what happens for example in U-Boot and Linux kernel - there are their own libc implementations - I would expect the same from Zephyr.

nashif:
> @nashif How did you do with OE-based tools before in regard to _SDATA symbol in the linker script?

We never had this issue before only now with ARC and this toolchain, I still need to look into it.



> As for newlib - do you really use it in Zephyr at all? Otherwise no need to spend time on our newlib fork.
> That's what happens for example in U-Boot and Linux kernel - there are their own libc implementations - I would expect the same from Zephyr.

We do have this as an option, some applications might request newlib, however by default right now we have minimal libc implementation that we use. So, we need to support newlib, but it is not a blocker for basic operation and zephyr usage.

nashif:
@abrodkin now that I have you here, any news on qemu for ARC? :)

abrodkin:
1. So what is an issue with this toolchain? I mean may I have more details on how this problem manifests etc?
2. Maybe not to bother with our newlib fork as we plan to improve upstream version in foreseeable future.

Issue #6, opened 2018-05-15
When support armv8 in toolchain - patch arm_cmse.h


galak:
The arm_cmse.h header uses typeof and it should really be using `__typeof__`

ioannisg:
FYI https://answers.launchpad.net/gcc-arm-embedded/+question/669076

galak:
We pull in the fix from upstream now, going to close.

PR #17, opened 2018-08-15
Update gcc-8.2 patch for arm_cmse.h c99 typeof fix

galak:
Pull in patch from upstream gcc that fixes use of typeof.


galak:
This fixes #6

PR #25, opened 2019-01-04
xtensa: fix toolchain and add overlays/hal

nashif:
Installs hal library as part of the toolchain. Currently only supports
newlib 2.2.


nashif:
> If in go.sh you are fetching xtensa_sample_controller.tar.gz why are we also committing a copy in overlays/xtensa_sample_controller.tar.gz?

sample_controller_linux.tgz is different from xtensa_sample_controller.tar.gz, it is based on it, but more xtools specific (overlay)

galak:
> sample_controller_linux.tgz is different from xtensa_sample_controller.tar.gz, it is based on it, but more xtools specific (overlay)

ah, yes I see that now.  We should document where the files we are committing are coming from.

nashif:
> ah, yes I see that now. We should document where the files we are committing are coming from.

Yes, will document all of that. All is based on reverse engineering how it was done in yocto based SDK, there are way too many steps generating those files and overlays that I really do not want to be scripting here, so it will be documented instead.

PR #26, opened 2019-01-08
xtensa: fix toolchain and add overlays/hal

galak:
Installs hal library as part of the toolchain. Currently only supports
newlib 2.2.


galak:
@nashif I updated your PR to use crosstool-ng to grab the newlib sources for xtensa from git instead of pulling in a tarball.

We might want to update overlays/README

PR #28, opened 2019-01-10
Prepare for adding tools to the generated SDK

nashif:
- Add needed scripts and CI to support host tools
- Remove companion tools from the toolchain build, those will come from the yocto based SDK


galak:
> in sdk-ng, old sdk will just provide the tools and they get composed in a CI step based on sdk-ng

Can we drop dtc from meta-zephyr-sdk and leave it building from sdk-ng?

nashif:
> Can we drop dtc from meta-zephyr-sdk and leave it building from sdk-ng?

why? we should build all tools in once place and only one time. Now we build DTC for each single toolchain and place it in the toolchain /bin. I do not think it makes sense to build tools in 2 different places.

galak:
> why? we should build all tools in once place and only one time. Now we build DTC for each single toolchain and place it in the toolchain /bin. I do not think it makes sense to build tools in 2 different places.

I guess my question is what's do we think our long term plan would be for building host tools like dtc & qemu?  Would we try to add those to crosstool-ng?

nashif:
> I guess my question is what's do we think our long term plan would be for building host tools like dtc, openocd, qemu, etc.? Would we try to add them to crosstool-ng?

depends on how complex it will get to do that and if crosstool-ng can give us portable enough tools that can work on multiple Linux systems without issues and if such tools can also can be upstreamed. I obviously would prefer that we only use one way for doing this so we can drop the yocto based tools completely. 

So until we can build all tools using xtools, I propose to stick to one method, and I have not seen any issues with building the tools using yocto yet.

Long-Long term I would even drop the tools completely and try to depend on what distros give us and in worst cases document how to build those :-)

PR #35, opened 2019-02-04
Disable small data for ARC

ruuddw:
Compile target compiler libraries without small data section.

Signed-off-by: Ruud Derwig Ruud.Derwig@synopsys.com


galak:
Why?  No issue with the change, would like to at least document in the PR why we made this change for ARC.

PR #47, opened 2019-02-22
update crosstool-ng and use new newlib 3.1.0.20181231

galak:
There isn't any difference between 3.1.0.20181231 and the official 3.1.0
release.  So drop our patch to use newlib 3.1.0 since now the upstream
crosstool-ng removed marking newlib 3.0.0.20180831 and 3.1.0.20181231 as
experimental.


galak:
This is mostly a maintenance patch, and push it to let it run through full sdk-ng CI.

PR #56, opened 2019-03-21
x86_64: adds config to build toolchain for x86_64

dcpleung:
This adds the config to build x86_64 toolchain with tuple
x86_64-multilib-linux-pc. This allows decoupling the builds from
the host toolchain.

This comes from the crosstool-ng's own x86_64-multilib-linux-gnu
sample config.

Signed-off-by: Daniel Leung <danielcp@gmail.com>


marc-hb:
BTW https://github.blog/2019-02-14-introducing-draft-pull-requests/

galak:
> can you add the x86_64 config to .shippable.yml in this PR so we actual try and build it.

also you need to update scripts/make_zephyr_sdk.sh for x86_64

dcpleung:
recheck

galak:
> recheck

not sure if recheck works on this repo.

dcpleung:
Just force pushed the commit without any changes to force CI to re-build.

marc-hb:
Closing/re-opening recently Just Worked for me.

galak:
I kicked the build of myself.

dcpleung:
> Closing/re-opening recently Just Worked for me.
> EDIT: force pushes work too, however github support for force pushes is hit and miss. Try for instance the "View Changes" button above.

Hm... maybe it's because there were no comments on the code itself?

> I kicked the build of myself.

Thanks.

marc-hb:
Force pushes are an afterthought in github, see https://github.com/zephyrproject-rtos/zephyr/pull/14444 for all the details. Github's preferred (and unusual) workflow is "fixup" commits.
(Gerrit for instance implements "magic" remote branches to automatically version rewritten history)
Off-topic, discuss in https://github.com/zephyrproject-rtos/zephyr/pull/14444 if needed.

dcpleung:
Should be okay now. I saw that x86_64.tar.bz2 was being uploaded to S3 in the log.

scripts/make_zephyr_sdk.sh
```diff
@@ -129,6 +130,13 @@ if [ -n "$file_gcc_riscv32" ]; then
   echo "echo \"\"" >>$setup
 fi
 
+if [ -n "$file_gcc_x86_64" ]; then
+  echo "tar -C \$target_sdk_dir -xf ./$file_gcc_x86_64 > /dev/null &" >> $setup
```
galak: can you update '-xf' to '-jxf`

dcpleung: On both? or just x86_64?

.shippable.yml
```diff
@@ -47,7 +48,11 @@ build:
           else
             echo "uploading ${TARGET}.tar.bz2 to ${S3_PATH}/";
             pushd build/output;
-            tar --exclude='build.log.bz2' -jcvf ../../${TARGET}.tar.bz2 *-zephyr-*;
+            if [ "${TARGET}" == 'x86_64' ]; then
```
galak: Why do we need to special case x86_64 here?

configs/x86_64.config
```diff
@@ -0,0 +1,20 @@
+CT_CONFIG_VERSION="2"
+CT_LOCAL_TARBALLS_DIR="${CT_PREFIX:-${HOME}/x-tools}/sources"
+# CT_PREFIX_DIR_RO is not set
+CT_PATCH_BUNDLED_LOCAL=y
+CT_LOCAL_PATCH_DIR="${CT_TOP_DIR}/../../patches"
+CT_ARCH_X86=y
+CT_MULTILIB=y
+CT_ARCH_64=y
+CT_TARGET_VENDOR="multilib"
```
galak: Why call this `multilib` instead of `zephyr`?

.shippable.yml
dcpleung: Because of the tuple, it is x86_64-multilib-linux-gnu

galak:
A few questions, going to make a 0.10.1 release with some bug fixes.

configs/x86_64.config
dcpleung: I think that *-zephyr-* implies that we are using newlib. However, this is using glibc instead because newlib does not fully support x32. It would result in linking errors for zephyr.elf.

galak: I don't see how or why `-zephyr-` would imply newlib.  I think it just means its from our SDK.

scripts/make_zephyr_sdk.sh
galak: Just x86_64, I updated the others upstream already.

configs/x86_64.config
dcpleung: Ok. I will change the tuple then.

scripts/make_zephyr_sdk.sh
dcpleung: Done.

.shippable.yml
dcpleung: After changing the vendor to zephyr, this is no longer needed.

configs/x86_64.config
dcpleung: Changed vendor to zephyr.

scripts/make_zephyr_sdk.sh
```diff
@@ -129,6 +130,13 @@ if [ -n "$file_gcc_riscv32" ]; then
   echo "echo \"\"" >>$setup
 fi
 
+if [ -n "$file_gcc_x86_64" ]; then
+  echo "tar -C \$target_sdk_dir -jxf ./$file_gcc_x86_64 > /dev/null &" >> $setup
```
marc-hb: `printf` would make quoting clearer and more robust, something like:
```
printf 'tar -C "%s" -jxf "%s" >/dev/null &'   '${target_sdk_dir}'  "./${file_gcc_x86_64}" >> "$setup"
```

marc-hb: Alternatively, a here-document would avoid the repetition:
```
cat <<EOFSETUP > "$setup"
tar -C "\${target_sdk_dir}" ...
etc.
EOFSETUP


```
Either would be better.

galak: Happy to take that as a followup cleanup patch

dcpleung:
The x86_64 build log in CI job #170 was empty, assuming it was never built. So re-triggering with a rebase to master...

nashif:
the logs are there, full of this:
```
[06:53] - 
[06:53] - 
[06:53] \ 
[06:53] \ 
[06:53] \ 
[06:53] \ 
[06:53] \ 
[06:53] \ 
[06:53] \ 
[06:53] \ 
[06:53] \ 
[06:53] \ 
[06:53] | 
[06:53] | 
[06:53] | 
[06:53] | 
[06:53] | 
[06:53] | 
[06:53] | 
[06:53] | 
[06:53] | 
[06:53] | 
[06:53] / 
[06:53] / 
[06:53] / 
[06:53] / 
[06:53] / 
[06:53] / 
[06:53] / 
[06:53] / 
[06:53] / 
[06:53] / 
```

@galak is this something new in crosstool-ng? is there a way to disable this for better console output?

galak:
> @galak is this something new in crosstool-ng? is there a way to disable this for better console output?

I haven't changed anything so not sure, maybe because its a linux build.

dcpleung:
The "bars" were because of the `CT_LOG_PROGRESS_BAR`. I have disabled it in latest patch, and those "bars" are no longer there.

configs/x86_64-zephyr-linux-gnu.config
```diff
@@ -0,0 +1,21 @@
...
+# CT_PREFIX_DIR_RO is not set
+CT_PATCH_BUNDLED_LOCAL=y
+CT_LOCAL_PATCH_DIR="${CT_TOP_DIR}/../../patches"
+# CT_LOG_PROGRESS_BAR is not set
+CT_ARCH_X86=y
+CT_MULTILIB=y
+CT_ARCH_64=y
+CT_TARGET_VENDOR="zephyr"
+CT_KERNEL_LINUX=y
+CT_BINUTILS_LINKER_LD_GOLD=y
```
galak: any particular reason to use GOLD?

dcpleung: Most of the config options here are copied straight from `crosstool-ng/samples/x86_64-multilib-linux-gnu/crosstool.config`. If you want, I can try without gold.

galak: Yes, lets try w/o gold.

dcpleung: Updated patch to remove gold.

galak:
As I think on this, why isn't this using newlib?

Seems like this should be more inline with the i586 toolchain.

dcpleung:
I looked into it for about a week or so. The issue I had was due to combination of crosstool-ng and newlib's build system. Somehow it resulted in libraries for exactly one architecture (even for gcc itself). Depending on how the "architectures to support list" is ordered, it would either result in all straight 64-bit libraries, or straight 32-bit libraries. So `zephyr.elf` would never be able to link correctly (as there is no x32), and bootstub might compile/link if there are 32-bit libraries laying around.

galak:
> I looked into it for about a week or so. The issue I had was due to combination of crosstool-ng and newlib's build system. Somehow it resulted in libraries for exactly one architecture (even for gcc itself). Depending on how the "architectures to support list" is ordered, it would either result in all straight 64-bit libraries, or straight 32-bit libraries. So `zephyr.elf` would never be able to link correctly (as there is no x32), and bootstub might compile/link if there are 32-bit libraries laying around.

I think we need a patch similar to the following for x86_64 to handle the multilibs:

https://github.com/zephyrproject-rtos/sdk-ng/blob/master/patches/gcc/8.3.0/0001-Enable-soft-fp-on-i586.patch

I'd rather focus on getting something like that working than having this franken toolchain for x86_64 that differs from the other SDK toolchains.

dcpleung:
Let me take a crack at it.

galak:
> Let me take a crack at it.

Posting a PR as a starting point:https://github.com/zephyrproject-rtos/sdk-ng/pull/72

The t-zephyr64 is a direct copy of t-linux64 right now, needs fixing

dcpleung:
> > Let me take a crack at it.
> 
> Posting a PR as a starting point:#72
> 
> The t-zephyr64 is a direct copy of t-linux64 right now, needs fixing

Thanks. I was missing the gcc patch when I tried last time. I will update this PR when I get it to work.

dcpleung:
Made some progress @ https://github.com/dcpleung/sdk-ng/commit/87ef3c70f1bb5cc8b83dc7811021b5cbf474b9bf . I was able to build GCC with `m32`, `mx32`, and `m64` support. However, newlib still is not building, as it is using `x86_64` files when doing `m32` build, resulting in incorrect `*.S` being pulled in with wrong registers.

dcpleung:
Updated the patch and now it's building newlib.

.shippable.yml
```diff
@@ -14,6 +14,7 @@ env:
       - TARGET=mips
       - TARGET=arc
       - TARGET=xtensa
+      - TARGET=x86_64-zephyr-elf
```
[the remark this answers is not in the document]
dcpleung: This is due to `scripts/make_zephyr_sdk.sh` only allowing one file name with substring `x86_64` inside `parse_toolchain_name()`. The host tool archive also has `x86_64` in its filename. The function would complain about multiple toolchains if the target and config file is simply `x86_64`.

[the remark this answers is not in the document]
galak: Ok, I think I understand, I guess we should open an issue to fix `scripts/make_zephyr_sdk.sh`.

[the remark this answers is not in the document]
dcpleung: Created an issue @ https://github.com/zephyrproject-rtos/sdk-ng/issues/74

patches/gcc/8.3.0/0001-initial-patch-for-x86_64.patch
```diff
@@ -0,0 +1,192 @@
+From 6037be9ef9636773e0702772ab86bb10e793f5be Mon Sep 17 00:00:00 2001
+From: Kumar Gala <kumar.gala@linaro.org>
+Date: Mon, 6 May 2019 15:55:48 -0500
+Subject: [PATCH 1/1] initial patch for x86_64
+
```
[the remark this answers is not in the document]
dcpleung: Added some explanation in latest patch.

patches/newlib/3.1.0.20181231/0002-Recognize-x86_64-zephyr-elf.patch
```diff
@@ -0,0 +1,559 @@
...
+
+diff --git a/config/dfp.m4 b/config/dfp.m4
+index 5b29089..00ea161 100644
+--- a/config/dfp.m4
++++ b/config/dfp.m4
+@@ -20,7 +20,7 @@ Valid choices are 'yes', 'bid', 'dpd', and 'no'.]) ;;
+ ],
+ [
+   case $1 in
+-    powerpc*-*-linux* | i?86*-*-linux* | x86_64*-*-linux* | s390*-*-linux* | \
```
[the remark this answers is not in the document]
dcpleung: Changed in latest patch.

PR #57, opened 2019-03-30
configs/all: add --enable-initfini-array to gcc configure

dcpleung:
The --enable-initfini-array configure flag forces GCC to use
init/fini arrays instead of just init/fini functions, as GCC
would choose to do functions depending on arch. With only
init/fini functions, apps built with coverage would not
dump any coverage data because __gcov_init() is never called.
With this flag, GCC will use init array which is required
for the correct operation of gcov_static_init().


galak:
Any reason i586 only?

nashif:
> Any reason i586 only?

we are only seeing this issue with x86 now.

galak:
> we are only seeing this issue with x86 now.

We should enable this across the board so all the arch's are consistent.

dcpleung:
Updated the PR with the same flag applied to all configs.

PR #69, opened 2019-05-01
riscv: Enable multilib support

galak:
Turn on multilib to support the various risc-v isa options.


galak:
Tested on both cores on vega and sifive hifive board.

Issue #75, opened 2019-05-16
asm code fails on cortex-m23


galak:
There's a bug if we build for cortex-m23 with GNU as.

Reported upstream:

https://sourceware.org/bugzilla/show_bug.cgi?id=24559

Once fixed will backport to the SDK.

galak:
Fixed with PR #77

PR #77, opened 2019-05-16
Cortex m23 asm fix

galak:
Address #75, making this draft for now til the fix gets committed upstream.


ioannisg:
This seems to solve the problems with compiling LDR immediate pseudo-instructions for ARMv8-M Baseline

PR #78, opened 2019-05-17
qemu: bump to 4.0.0 release

galak:
Updated nios2 patch to build with qemu 4.0.0


ioannisg:
@galak I am wondering: is there more we need to do, to move to 4.0.0 qemu release?

galak:
> @galak I am wondering: is there more we need to do, to move to 4.0.0 qemu release?

We need to test this in CI to see how it works.

ioannisg:
> 
> 
> > @galak I am wondering: is there more we need to do, to move to 4.0.0 qemu release?
> 
> We need to test this in CI to see how it works.

How will we do that? Do we have a standard process, @galak ?

ioannisg:
I can report, that I managed to run the kernel, and user-mode test suite on qemu4 (an521) with just a couple of tests failing. I am investigating these failures at the moment.

galak:
> How will we do that? Do we have a standard process, @galak ?

We don't have a standard process.  One way is to do it locally, shouldn't be too bad to get a new SDK build and run/test all the QEMU targets against it.

PR #88, opened 2019-08-05
qemu: enable riscv64-softmmu

galak:


configs/riscv64.config
```diff
@@ -0,0 +1,29 @@
...
+CT_LOCAL_TARBALLS_DIR="${CT_PREFIX:-${HOME}/x-tools}/sources"
+# CT_PREFIX_DIR_RO is not set
+CT_PATCH_BUNDLED_LOCAL=y
+CT_LOCAL_PATCH_DIR="${CT_TOP_DIR}/../../patches"
+# CT_LOG_PROGRESS_BAR is not set
+CT_ARCH_RISCV=y
+CT_MULTILIB=y
+CT_ARCH_64=y
+CT_ARCH_ARCH="rv32ima"
+CT_ARCH_ABI="ilp32"
```
olofj: Hmm this doesn't look right. Should be lp64 and rv64ima above, I suspect.

Also, might as well do a multiarch riscv that takes -m32/-m64 in the same toolchain, right?

galak: The ARCH & ABI are just setting the default if what the toolchain will build if -mabi and -march aren't specified.  I think we are getting a multiarch build that supports 32/64, from what I can tell on RISC-V that is done via the -mabi choice as there isn't a documented -m32/-m64 option.

PR #102, opened 2019-08-21
turn on some newlib enhc (LTO)

galak:


stephanosio:
NOTE: If we are ever going to apply this patch, once #152 is merged, we should apply it to both normal and nano variants:

```
+CT_LIBC_NEWLIB_EXTRA_SECTIONS=y
+CT_LIBC_NEWLIB_LTO=y

+CT_LIBC_NANO_NEWLIB_EXTRA_SECTIONS=y
+CT_LIBC_NANO_NEWLIB_LTO=y
```

galak:
@stephanosio do you mind looking at this and seeing if there's a code size difference if we enable these options?

stephanosio:
> @stephanosio do you mind looking at this and seeing if there's a code size difference if we enable these options?

Will do.

stephanosio:
```
-rw-r--r-- 1 stephanos stephanos 5.2M Dec 10 07:59 libc.a
-rw-r--r-- 1 stephanos stephanos 8.6M Dec 10 07:57 libc_nano.a
```

The `nano` variant seems bigger because it is being built with debug information and LTO adds extra debug sections that are not removed in the build script.

A quick attempt to strip the libraries results in the following error:

```
stephanos@DOMINUS:~/Dev/archive/zephyr-sdk-ng_newlib-enh-20191210_0141/arm-zephyr-eabi/lib$ ../../bin/arm-zephyr-eabi-strip --strip-debug libc_nano.a
../../bin/arm-zephyr-eabi-strip: stJmdq3d/lib_a-_Exit.o: plugin needed to handle lto object
../../bin/arm-zephyr-eabi-strip: stJmdq3d/lib_a-__adjust.o: plugin needed to handle lto object
../../bin/arm-zephyr-eabi-strip: stJmdq3d/lib_a-__atexit.o: plugin needed to handle lto object
../../bin/arm-zephyr-eabi-strip: stJmdq3d/lib_a-__call_atexit.o: plugin needed to handle lto object
...
```

This seems to be because we have not enabled LTO support when building binutils.

In general, I do not think we are not ready for LTO at this time, especially considering that the GNU ARM Embedded toolchain does not enable or support it. There is also some inherent risk associated with it, as LTO is known to often break things and was never comprehensively tested for the ARM EABI.

galak:
What about the extra sections?  Will are going to notice this when we try and link with newlib.  Ideally we see a footprint reduction as the linker an remove stuff we dont use.

stephanosio:
> What about the extra sections? Will are going to notice this when we try and link with newlib. Ideally we see a footprint reduction as the linker an remove stuff we dont use.

I created a separate branch/PR #157 for that. I will add test results there.

stephanosio:
Closing stale PR.

PR #106, opened 2019-08-27
Enable nano.spec layout

galak:
Pull in a patch to crosstool-ng that adds a config option to match the
file layout as expected by nano.spec.  Enable this option so that we
match behavior with other toolchain builds (ARM GCC embedded, various
cross distro toolchains, etc).  This paves the way for us to support a
different variant build of newlib (a fuller version that doesn't enable
the nano options for example).


stephanosio:
Force pushing to resolve merge conflicts.

stephanosio:
> Will hold on this until we have a solution on building 2 newlib variants (nano and full).

working on a PR to crosstool-ng for that at the moment.

PR #107, opened 2019-09-11
xtensa: build toolchains in their own directories

dcpleung:
Xtensa are configurable processors where each SoC requires its own custom toolchain. However, by default, every single one is built with prefix xtensa-zephyr-elf. When building multiple ones, they are going to overwrite each other in the output directory. In order to support multiple Xtensa-based SoCs, their toolchains are now built under their own directory.

This is in preparation for introducing additional Xtensa toolchains.


galak:
Closing this, will just merge #111

PR #111, opened 2019-09-26
Add configs to build Xtensa toolchains for Intel S1000 and for Audio DSP on Apollo Lake

dcpleung:
This adds two configurations to build Xtensa toolchains for Intel S1000 (board intel_s1000_crb) and for the audio DSP on Apollo Lake.


go.sh
```diff
@@ -68,11 +68,10 @@ fi
 
 CT_NG=${SDK_NG_HOME}/bin/ct-ng
 
-
 mkdir -p build
 cd build
-export CT_PREFIX=`pwd`/output
-mkdir -p ${CT_PREFIX}/sources
+export OUTPUT_DIR=`pwd`/output
```
galak: Why was this change needed?

dcpleung: This is to group all the Xtensa toolchains under one directory (see line 90-99 below). The intention is to have the same directory name as SoC name under `build/output/xtensa/${SOC}` in Zephyr cmake.

dcpleung:
> Why do we need the GDB change? Why can't we use GDB 8.2 on xtensa_intel_apl_adsp.config and xtensa_intel_s1000?

The `intel_s1000` requires binary only tool `xt-ocd` from Cadence for debugging and it only works with GDB 7. From what I can tell so far, the data packet detailing registers from `xt-ocd` cannot be parsed correctly in GDB 8 due to size mismatch.

As for `intel_apl_adsp`, it mirrors what the SOF project is using for GDB (@ https://github.com/thesofproject/crosstool-ng/blob/sof-gcc8.1/config-apl-gcc8.1-gdb8.1).

galak:
> As for `intel_apl_adsp`, it mirrors what the SOF project is using for GDB (@ [thesofproject/crosstool-ng:config-apl-gcc8.1-gdb8.1@`sof-gcc8.1`](https://github.com/thesofproject/crosstool-ng/blob/sof-gcc8.1/config-apl-gcc8.1-gdb8.1) [](/thesofproject/crosstool-ng/blob/HEAD@{2019-10-03T18:10:19Z}/config-apl-gcc8.1-gdb8.1)).

I'd prefer we keep such differences to a minimum, so for `intel_apl_adsp` I think we should use current gdb.

dcpleung:
> > As for `intel_apl_adsp`, it mirrors what the SOF project is using for GDB (@ [thesofproject/crosstool-ng:config-apl-gcc8.1-gdb8.1@`sof-gcc8.1`](https://github.com/thesofproject/crosstool-ng/blob/sof-gcc8.1/config-apl-gcc8.1-gdb8.1) ).
> 
> I'd prefer we keep such differences to a minimum, so for `intel_apl_adsp` I think we should use current gdb.

I have update the config file for `intel_apl_adsp`.

PR #122, opened 2019-10-12
x86_64: update GCC 9.2.0 patch file and default march

dcpleung:
* Ran autoreconf on library projects to support the x86_64-zephyr-elf tuple.
* Remove the check for multilib in configure as we are now hard-coding the MULTILIB_OPTIONS in t-zephyr64. The previos attempt was using the Linux version of t-linux64. That is targeting shared library placement for live system. Since Zephyr only needs the static libraries for linking purpose only, there is no need to put them under /lib or /lib64. So simplify it a bit.
* Set m64 as the default for multilib so it won't build 64-bit libraries twice (one under /lib, and one under /lib/64).
* Add LINK_SPEC to specify default binary format for final linking. This helps with compiler flag check with cmake as it does not know what format to use, and the default elf_x86_64 was being used to link both 32 and x32 targets. The check would fail as as the compiler would always try to link 64-bit libgcc.a even for 32-bit or x32 targets.
* Skip the need for crt0.o as there is no crt0.o for x86_64 targets. The absence of this file also causes compiler flag check failures.
* Set the default architecture for both 32-bit and 64-bit targets.

### DIFF
--- a/configs/x86_64-zephyr-elf.config
+++ b/configs/x86_64-zephyr-elf.config
@@ -18,8 +18,8 @@ CT_LIBC_NEWLIB_LITE_EXIT=y
 # CT_LIBC_NEWLIB_WIDE_ORIENT is not set
 CT_LIBC_NEWLIB_NANO_MALLOC=y
 CT_LIBC_NEWLIB_NANO_FORMATTED_IO=y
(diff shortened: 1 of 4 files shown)


galak:
Can we use the x86_64 compiler for i586?

dcpleung:
> Can we use the x86_64 compiler for i586?

Not yet. I am looking into it for awhile but getting lots of kernel panic in sanitycheck.

PR #124, opened 2019-10-20
meta-zephyr-sdk: build custom SeaBIOS for QEMU

dcpleung:
This adds a receipe to build SeaBIOS for use with QEMU,
which includes a patch to avoid clearing screen and
disabling line wrapping in serial console. This helps
with looking at emulator outputs (from both running
locally and sanitycheck) as clearing screen and disabling
line wrapping affect developer terminal, requiring
running 'reset' to make it useful again.

The config files comes from QEMU.

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/11717

The code can be found https://github.com/dcpleung/seabios/tree/zephyr


galak:
@nashif should we create a repo for SeaBIOS under Zephyr?

We need some location for the changes to SeaBIOS (even as patches in the sdk repo) that aren't just in @dcpleung git repo.

nashif:
> @nashif should we create a repo for SeaBIOS under Zephyr?
> 
> We need some location for the changes to SeaBIOS (even as patches in the sdk repo) that aren't just in @dcpleung git repo.

you have the patch in the PR already which will go into the SDK

meta-zephyr-sdk/recipes-devtools/qemu/files/0002-pc-bios-Update-SeaBIOS-to-skip-clearing-screen.patch

Does the patch justify a git repo? We have other patches under meta-zephyr-sdk..

galak:
> you have the patch in the PR already which will go into the SDK
> 
> meta-zephyr-sdk/recipes-devtools/qemu/files/0002-pc-bios-Update-SeaBIOS-to-skip-clearing-screen.patch

This patch is replacing the binary files. Not the actual source change to SeaBIOS.
 
> Does the patch justify a git repo? We have other patches under meta-zephyr-sdk..

Ideally we'd build SeaBIOS as part of `meta-zephyr-sdk`.

nashif:
> This patch is replacing the binary files. Not the actual source change to SeaBIOS.

oh ok, did not see that :) 
and this binary is built by @dcpleung, I see now.

> Ideally we'd build SeaBIOS as part of meta-zephyr-sdk.

Right, hopefully it will not pull all kind of dependencies.

But that does not mean we should host the git repo for seabios, we can build it in the SDK with the patches from @dcpleung, right?

dcpleung:
I thought about building with the sdk, but I am not too familiar with Yocto and have no idea how to use the custom-built SeaBIOS in QEMU build. Or we could build SeaBIOS binaries and put them into different directory under sysroot, and point to those binaries in the build system.

galak:
> I thought about building with the sdk, but I am not too familiar with Yocto and have no idea how to use the custom-built SeaBIOS in QEMU build. Or we could build SeaBIOS binaries and put them into different directory under sysroot, and point to those binaries in the build system.

I wonder if we can pull something like the following in:

https://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization/tree/recipes-extended/seabios/seabios_1.9.0.bb?h=jethro

galak:
I newer version:

https://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization/tree/recipes-extended/seabios/seabios_1.11.2.bb

dcpleung:
Updated the PR to build SeaBIOS binaries and have QEMU pull them in before building. Used MD5SUM to verify files are actually the custom built binaries.

pfalcon:
Thanks for working to resolve this very annoying issue. I actually saw even more weird effects of this SeaBIOS screen initialization, specifically: after running such a QEMU session, "git add -p" (interactive add chunk by chunk) stopped working, specifically it just quit on first interactive prompt, kinda as if it was run as `git add -p </dev/null`, i.e. it saw EOF of reading stdin. What's even more weird is that usually ill effects like disabling line wrapping can be cleared with the "reset" command, but that stdin issue survived across it. Literally, that particular terminal session was hosed, and you get to open a new window `git add -p`/`git co -p`.

I can't say it happened 100% after each QEMU run (and I don't use `git add -p`/`git co -p` like every half an hour to notice it), but I see it regularly. Bottom line @dcpleung, as far as can tell, your patch just disables any terminal initialization in QEMU altogether, but if not, and you know that something does weird things to stdin config too, may be a good idea to patch it out too. Thanks.

dcpleung:
Hm... I have never seen any issue with git after running QEMU. Let me see if I can reproduce it on my end.

dcpleung:
CI is passing now. Should be good to go.

PR #125, opened 2019-10-23
meta-zephyr-sdk: hidapi-libraw: add libusb1 as dependency

dcpleung:
Building tools always fails on first pass where hidapi-libraw
cannot find libusb-1.0 during configure. So explicitly add
libusb1 as dependency for hidapi-libraw so it won't fail.


galak:
Just curious how the issue shows up and how it is that we are still able to generate the tools?

No issue with the change, just curious to understand.

dcpleung:
It always happens when I run `go.sh tools` or running `meta-zephyr-sdk-build.sh tools` manually, but it only happens the very first time. Right now, it works because `meta-zephyr-sdk-build.sh` is run three times inside `go.sh`. From what I can tell, libusb gets pulled in by another package (possibly openocd by grepping), so running build second time works since libusb is already there. If there are not other dependency issues, we can remove the 2nd and 3rd call to build.

PR #170, opened 2020-01-10
arm: strip enable

galak:


stephanosio:
Closing as this is not relevant.

Note that `CT_STRIP_HOST_TOOLCHAIN_EXECUTABLES` already [takes care of stripping all the host toolchain binaries (including binutils and gcc)](https://github.com/crosstool-ng/crosstool-ng/blob/5659366bf62b5555bf914b5f55e8a01c92d6c6f1/scripts/build/internals.sh#L77-L89).

The only effect of `CT_STRIP_TARGET_TOOLCHAIN_EXECUTABLES` is on the gcc/libstdc++ build process in which [`install-strip` is passed where the host gcc executables](https://github.com/crosstool-ng/crosstool-ng/blob/c969be1880f99b6da4d98afd7bda641b7868a717/scripts/build/cc/gcc.sh#L1180-L1184) and the target libstdc++ libraries are stripped.

This is unnecessary/undesirable because:

1. Host gcc executables are already stripped by `CT_STRIP_TARGET_TOOLCHAIN_EXECUTABLES`
2. Target libraries (including newlib and libstdc++) are already stripped by [`CT_STRIP_TARGET_TOOLCHAIN_LIBRARIES`](https://github.com/zephyrproject-rtos/sdk-ng/blob/master/patches/0005-Allow-stripping-target-toolchain-libraries.patch) (also note that this stripping procedure requires special handling to preserve the `.debug_frame` section which is required for stack unwinding by various debuggers. Using `.eh_frame` is not an option because that actually gets included into the `text` and affects the firmware binary size.

PR #172, opened 2020-01-15
Openocd 20200114

galak:


erwango:
@galak Is that possible to include next 4 commits from openocd tree (251eb035fc72c8fe6348fa21d8c8d8faa9902988) ?

It allows to flash nucleo_wb55rg (much faster than pyocd, so better for sanitycheck) and adds support for few additional L4 targets.

galak:
Updated this to a snapshot based on Jan 16, 2020 which should pull in the changes you mentioned.

PR #184, opened 2020-02-05
Revert "gdb: Build gdb static" as it disables python support

galak:
The attempt to build python statically was to address an issue with
python library being needed, however this ends up breaking python
support so lets revert the commit and we'll document the need to have
python3.6.

This reverts commit 6ad840ac76b699a4abf54fe08145b0b1986471ce.


galak:
Need to add a doc update for this.

carlescufi:
cc @joerchan

joerchan:
Thank you!

galak:
Documenting this here:

On fedora if python3.6 isn't installed we get:

```
arm-zephyr-eabi-gdb: error while loading shared libraries: libpython3.6m.so.1.0: cannot open shared object file: No such file or directory
```

fixed by `sudo dnf install python36`

PR #191, opened 2020-02-26
xilinx_qemu: Fix WFI instruction for icount mode

stephanosio:
The Xilinx QEMU deliberately breaks the WFI (wake up from interrupt)
instruction when icount mode is used, possibly for performance and
cosimulation stability reasons, and this causes the Zephyr CI tests,
that expect the WFI instruction to function properly, to fail.

This commit adds a patch to enable the CPU halt behaviour of the WFI
instruction in the icount mode, since accurate simulation behaviour
takes precedence over performance for CI testing, and the Zephyr will
not use the cosimulation feature.

This PR addresses the issue identified in https://github.com/zephyrproject-rtos/zephyr/pull/22904#issuecomment-591209160.


stephanosio:
cc @wentongwu

galak:
@stephanosio is this really 0.12.0 material or would this be reasonable for 0.11.3 release?

stephanosio:
> @stephanosio is this really 0.12.0 material or would this be reasonable for 0.11.3 release?

@galak Since this is a rather minor change, I suppose 0.11.3 would be reasonable.

PR #195, opened 2020-03-07
xtensa: intel_s1000: update to GDB 8.3.1

### LINKED ISSUE #188: Upgrade GDB on xtensa_intel_s1000 to 8.3
See comments in PR #187 about testing of GDB 8.3.

dcpleung:
The intel_s1000 SoC (and the corresponding intel_s1000_crb board)
uses Tensilica's own xt-ocd for debugging, which works well with
Tensilica's own xt-gdb but based on an older version. To make
upstream GDB 8.3 work with xt-ocd, the following modifications
are needed:

() Introduce an option to use the remote target register number
   instead of the sequential number as defined in xtensa-config.c.
   This affects read/write of registers as xt-ocd maps registers
   to their target number. QEMU, on the other hand, uses
   the sequential number. Therefore only turns on this option
   when building GDB for intel_s1000.

() Registers that cannot be retrieved are marked unavailable
   instead of erroring out. There are priviledged registers which
   can only be retrieved with custom protocol used by xt-gdb.
   Upstream GDB does not support this. So instead of error out
   and stopping register retrieval, simply mark those registers
   as unavailable and continue reading remaining registers.

() Add an option to not use 'g' packet. When GDB sends a 'g'
   packet to retrieve a group of registers, xt-ocd does not
   produce responses expected by upstream GDB, and GDB simply
   goes into an infinite loop hoping a correct response would
   eventually come. So this option is introduced to mark all
   as not in 'g' packet so GDB would use 'p/P' packets to
   read/write registers.

Fixes #188


dcpleung:
These are more like hacks to get upstream GDB to work with `xt-ocd`. I don't know if upstream would want these. Most people using Tensilica's tools would probably be using their `xt-gdb` anyway. I can try upstreaming these and see how it goes.

PR #199, opened 2020-04-07
cmake: Introduce CMake package for Zephyr SDK

tejlmand:
This commit introduces CMake package for Zephyr SDK.

This allows users to install Zephyr SDK at the following locations:
```
$HOME/zephyr-sdk*
$HOME/.local/zephyr-sdk*
$HOME/.local/opt/zephyr-sdk*
/opt/zephyr-sdk*
/usr/zephyr-sdk*
/usr/local/zephyr-sdk*

```
and have the Zephyr build system automatically detect the installation.

The * indicates that it is allowed to install the Zephyr SDK in a
location such as: /opt/zephyr-sdk-0.11.0 and it will still be found by
CMake.

Using CMake package version check further ensures that multiple SDK
versions can be installed, and CMake will select the version closets to
the required.

For example, requiring Zephyr SDK 0.11.1 in a system where the following
versions are installed:
- /opt/zephyr-sdk-0.10.0 : 0.10.0  (Not matching)
- /opt/zephyr-sdk-0.11.1 : 0.11.1  (First matching - will be used)
- /opt/zephyr-sdk-0.12.0 : 0.12.0  (Matching)

in this case, both 0.11.1 and 0.12.0 are matching the requirement, but
0.11.1 will be chosen as it is closest to the required version.

The files from Zephyr repo: `host-tools.cmake`, `generic.cmake`, and
`target.cmake` are also moved to the SDK repository, as it allows the
Zephyr build system to include the correct cmake files, even when using
a future SDK release with older Zephyr, as long as the Zephyr SDK is
backward compatible.

----
Additional info:

Today, when doing a bisect or jumping to an LTS release, for example Zephyr 1.14.1, but having a Zephyr-SDK `0.11.1` will give an error similar to this:
```
tora@ubuntu /p/g/n/z/s/hello_world> rm -rf build; cmake -GNinja -DBOARD=nrf51_pca10028 -Bbuild .
Zephyr version: 2.1.0
-- Found PythonInterp: /usr/bin/python3 (found suitable version "3.6.8", minimum required is "3.4") 
-- Selected BOARD nrf51_pca10028
CMake Error at /projects/github/ncs/zephyr/cmake/toolchain/zephyr/host-tools.cmake:57 (include):
  include could not find load file:

    /projects/github/ncs/zephyr/cmake/toolchain/zephyr/0.11/host-tools.cmake
Call Stack (most recent call first):
  /projects/github/ncs/zephyr/cmake/host-tools.cmake:3 (include)
  /projects/github/ncs/zephyr/cmake/app/boilerplate.cmake:443 (include)
  CMakeLists.txt:5 (include)


-- Found west: /home/tora/.local/bin/west (found suitable version "0.7.2", minimum required is "0.6.0")
CMake Error at /projects/github/ncs/zephyr/cmake/toolchain/zephyr/generic.cmake:7 (include):
  include could not find load file:

    /projects/github/ncs/zephyr/cmake/toolchain/zephyr/0.11/generic.cmake
Call Stack (most recent call first):
  /projects/github/ncs/zephyr/cmake/generic_toolchain.cmake:64 (include)
  /projects/github/ncs/zephyr/cmake/app/boilerplate.cmake:459 (include)
  CMakeLists.txt:5 (include)


-- Loading /projects/github/ncs/zephyr/boards/arm/nrf51_pca10028/nrf51_pca10028.dts as base
CMake Error at /projects/github/ncs/zephyr/cmake/dts.cmake:130 (message):
  command failed with return code: No such file or directory
Call Stack (most recent call first):
  /projects/github/ncs/zephyr/cmake/app/boilerplate.cmake:460 (include)
  CMakeLists.txt:5 (include)
```
or
```
tora@ubuntu /p/g/n/z/s/hello_world> rm -rf build; cmake -GNinja -DBOARD=nrf51_pca10028 -Bbuild .
Zephyr version: 1.14.1
-- Found PythonInterp: /usr/bin/python3 (found suitable version "3.6.8", minimum required is "3.4") 
-- Selected BOARD nrf51_pca10028
CMake Error at /projects/github/ncs/zephyr/cmake/toolchain/zephyr/host-tools.cmake:47 (include):
  include could not find load file:

    /projects/github/ncs/zephyr/cmake/toolchain/zephyr/0.11/host-tools.cmake
Call Stack (most recent call first):
  /projects/github/ncs/zephyr/cmake/host-tools.cmake:3 (include)
  /projects/github/ncs/zephyr/cmake/app/boilerplate.cmake:382 (include)
  CMakeLists.txt:5 (include)


-- Found west: /home/tora/.local/bin/west (found suitable version "0.7.2", minimum required is "0.5.6")
CMake Error at /projects/github/ncs/zephyr/cmake/toolchain/zephyr/generic.cmake:7 (include):
  include could not find load file:

    /projects/github/ncs/zephyr/cmake/toolchain/zephyr/0.11/generic.cmake
Call Stack (most recent call first):
  /projects/github/ncs/zephyr/cmake/generic_toolchain.cmake:57 (include)
  /projects/github/ncs/zephyr/cmake/app/boilerplate.cmake:398 (include)
  CMakeLists.txt:5 (include)


-- Loading /projects/github/ncs/zephyr/boards/arm/nrf51_pca10028/nrf51_pca10028.dts as base
-- Overlaying /projects/github/ncs/zephyr/dts/common/common.dts
CMake Error at /projects/github/ncs/zephyr/cmake/dts.cmake:126 (message):
  command failed with return code: No such file or directory
Call Stack (most recent call first):
  /projects/github/ncs/zephyr/cmake/app/boilerplate.cmake:399 (include)
  CMakeLists.txt:5 (include)


-- Configuring incomplete, errors occurred!

```

This PR removes the need for setting `ZEPHYR_SDK_INSTALL_DIR` or `ZEPHYR_TOOLCHAIN_VARIANT` if installing Zephyr SDK under one the following locations:
- `${HOME}/zephyr-sdk*`
- `/opt/zephyr-sdk*`
- `/usr/zephyr-sdk*`
- `/usr/local/zephyr-sdk*`
where the `*` denotes that anything can be added after the beginning, that is:
`/opt/zephyr-0.11.1` and `/opt/zephyr-sdk` are both valid.
This means that in Zephyr repo the following code is added, and Zephyr uses `find_package(Zephyr-sdk VERSION)` to lookup Zephyr.

This has the following advantages:
1. Installing multiple Zephyr sdk's without changing environment
1. Source the additional cmake files required from the Zephyr SDK instead, making it possible to use a future Zephyr SDK with older build (the SDK decides if it is backward compatible)
1. Zephyr build will select the best matching version, to the minimum required.

As example, having a newer Zephyr SDK `0.12.0` but no `${MAJOR_MINOR}` version in includes will now work, for example:
```
 find_package(Zephyr-sdk 0.11.1)
```
will result in `0.12.0` being used, if that is only Zephyr SDK installed (and is compatible):
```
-- Found toolchain: zephyr (/opt/zephyr-sdk-0.12.0)
```
but let's say user only has `0.13.0` and this one is not compatible with `0.11.1` requirement, then the following message is printed:
```
  The SDK version you are using is not supported, please update your SDK.

  You need at least SDK version 0.11.1.

  You have version(s):

    0.10.0 (/opt/zephyr-sdk-0.10.0)
    0.13.0 (/opt/zephyr-sdk-0.13.0)

  The new version of the SDK can be downloaded from:
  
  https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v0.11.1/zephyr-sdk-0.11.1-setup.run
```
which is much better user experience thanb todays cryptic message.


tejlmand:
@galak I cannot assign anyone as reviewer, could you please add relevant people.

I updated the `go.sh`, `template_dir`, and `.shipable.yml` but I can see the `make_zephyr_sdk.sh` is using `toolchain` folder, however, I don't find the location where shippable or `go.sh` moves build output to that folder automatically, so i'm unsure if I have missed a files that needs to be updated.

I create a `cmake.tar.bz2` in `go.sh` which contains the package files needed in the install folder.

galak:
DEV-REVIEW (april 16):
* @nashif - need to maintain existing functionality and documenting
* @galak - look at adding .github workflow that verifies this works using both the cmake package method and env method [as a followup]

cmake/Zephyr-sdkConfig.cmake
```diff
@@ -0,0 +1,20 @@
...
+# Those are Zephyr variables used.
+set(SDK_VERSION ${Zephyr-sdk_VERSION})
+set(SDK_MAJOR_MINOR ${Zephyr-sdk_VERSION_MAJOR}.${Zephyr-sdk_VERSION_MINOR})
+set(SDK_VERSION ${Zephyr-sdk_VERSION_MAJOR}.${Zephyr-sdk_VERSION_MINOR}.${Zephyr-sdk_VERSION_PATCH})
+set(SDK_MAJOR_MINOR_MICRO ${SDK_VERSION})
+
+get_filename_component(ZEPHYR_SDK_INSTALL_DIR ${CMAKE_CURRENT_LIST_DIR}/.. ABSOLUTE)
+set(ZEPHYR_SDK_INSTALL_DIR ${ZEPHYR_SDK_INSTALL_DIR})
+set(ZEPHYR_TOOLCHAIN_VARIANT zephyr)
+set(TOOLCHAIN_VENDOR zephyr)
```
carlescufi, line 15: Do we still need this?

cmake/Zephyr-sdkConfig.cmake
```diff
@@ -0,0 +1,20 @@
...
+set(TOOLCHAIN_ARCH x86_64)
```
carlescufi, line 16: Do we still need this?

tejlmand: This is used here:
https://github.com/zephyrproject-rtos/sdk-ng/blob/c90442fc6875ac34167c0a2948893ec28d93ed4e/cmake/zephyr/host-tools.cmake#L3

but as this file is now moved out of Zephyr repo, and into Zephyr SDK, there is no reason for going through a variable, as it is not used elsewhere in Zephyr.

So will fix.

cmake/Zephyr-sdkConfig.cmake
tejlmand: I was used for `0.10` but no longer for `0.11`, so will remove.
Unless someone comes up with a reason to keep it.

cmake/Zephyr-sdkConfig.cmake
galak: Is this fix on the sdk-ng side or zephyr side?

tejlmand: on sdk-ng side.
Fixed now.

cmake/Zephyr-sdkConfig.cmake
tejlmand: fixed

tejlmand:
as per request, https://github.com/zephyrproject-rtos/zephyr/pull/24143#discussion_r411329532, `$HOME/bin/zephyr-sdk[-x.y.z]` is now added to list of install folders that is default handled.

tejlmand:
I have received a small question if this could also be registered in CMake package registry to allow complete freedom of installation.

Will investigate a bit, as this should would need to be part of the install script, but I prefer to avoid execution of CMake in the install script.

galak:
> I have received a small question if this could also be registered in CMake package registry to allow complete freedom of installation.
> 
> Will investigate a bit, as this should would need to be part of the install script, but I prefer to avoid execution of CMake in the install script.

Yeah, I'm not in favor of having to run cmake in the install script.

tejlmand:
> > I have received a small question if this could also be registered in CMake package registry to allow complete freedom of installation.
> > Will investigate a bit, as this should would need to be part of the install script, but I prefer to avoid execution of CMake in the install script.
> 
> Yeah, I'm not in favor of having to run cmake in the install script.

Made a commit that is not using CMake, but registering the Zephyr SDK in the CMake package registry in same way as CMake does it.
https://github.com/zephyrproject-rtos/sdk-ng/pull/199/commits/8f443d444dd8b7c7a13ab82d588e994798591130

tejlmand:
> Are the file names Zephyr-c.cmake and Zephyr-sdkConfig.cmake special? Can we have them be zephyr-Zephyr-sdkConfig.cmake?

Zephyr-c.cmake: I assume you mean Zephyr-sdkConfigVersion.cmake and Zephyr-sdkConfig.cmake ?
And yes, those names are special.

In order for CMake to be able to identify a package, then it will look for the files:
`<package-name>Config.cmake` and `<package-name>ConfigVersion.cmake`.
So if we want to find the SDK using:
```
find_package(Zephyr-sdk)
```
then the files must be named: `Zephyr-sdkConfigVersion.cmake` and `Zephyr-sdkConfig.cmake`

Note: the files can also be named `zephyr-sdk-config.cmake` and `zephyr-sdk-config-version.cmake`
That is all lower case, and with dash as separator.
See: https://cmake.org/cmake/help/v3.17/command/find_package.html#full-signature-and-config-mode

So that's the two versions we can choose from, and I just picked the version I generally have seen mostly used.

daor-oti:
Is there any way we could make it possible to add an optional custom sdk install path?
This would allow for auto resolve of sdk version, even in situations where the sdk cannot be installed at any of the default locations.
Perhaps the optional install search path could be defined in the .zephyrrc as ZEPHYR_SDK_SEARCH_PATH or something along those lines?
The search path entry in the .zephyrrc could be generated during the sdk installation from the install path specified.

scripts/template_dir
```diff
@@ -188,6 +224,39 @@ if [ -n "$(echo $target_sdk_dir|grep ' ')" ]; then
 	exit 1
 fi
 
+# Verify that the install forlder is among the default search paths for CMake.
```
galak, line 227: forlder -> folder

tejlmand:
> do_cmake_package isn't invoked if I install in a default location. This seems like a bug??

No, this is correct behavior. The default locations are those we will always check for a Zephyr SDK, hence there is no need to register anything in the CMake package registry.

Only if user decides to install outside of those locations, then the only way we can find the SDK is by an entry in the package registry.

scripts/template_dir
tejlmand: fixed

daor-oti:
@tejlmand Thank you for the explanation and that is great, then no changes will be necessary to support installing to other locations and keeping the auto resolve of the sdk version, perfect.

tejlmand:
> @tejlmand Thank you for the explanation and that is great, then no changes will be necessary to support installing to other locations and keeping the auto resolve of the sdk version, perfect.

@daor-oti Added this description in the updated docs:

>   `ZEPHYR_SDK_INSTALL_DIR` can also be used for pointing to a folder
>   containing multiple Zephyr SDKs, allowing for automatic toolchain selection,
>   for example: ``ZEPHYR_SDK_INSTALL_DIR=/company/tools``
>
>   * ``/company/tools/zephyr-sdk-0.11.3``
>   * ``/company/tools/zephyr-sdk-a.b.c``
>   * ``/company/tools/zephyr-sdk-x.y.z``
>
>   this allow Zephyr to pick the right toolchain, while allowing multiple Zephyr
>   SDKs to be grouped together at a custom location.

in this commit: https://github.com/zephyrproject-rtos/zephyr/pull/24143/commits/a69ba375bb296624e5714a26100d14d1ffd13d4c

daor-oti:
@tejlmand That is great, thanks. Now it is clear that the ZEPHYR_SDK_INSTALL_DIR can also be used this way.

PR #202, opened 2020-04-10
gdb: Force using Python 2.7

### LINKED ISSUE #165: error when running ARC GDB in SDK on Ubuntu 19.10
`/opt/zephyr-sdk/arc-zephyr-elf/bin/arc-zephyr-elf-gdb: error while loading shared libraries: libpython2.7.so.1.0: cannot open shared object file: No such file or directory`

Workaround:

```
 $ sudo apt install libpython2.7
```

Workstation setup instructions may need to be updated, or SDK build parameters adjusted.
I haven't tested whether this problem is specific to ARC's GDB in the SDK, or other platforms.

### LINKED ISSUE #196: dependency on python 3.6 headers breaks on some Debian/Ubuntu versions
/home/daniel/zephyr-sdk-0.11.2/arm-zephyr-eabi/bin/arm-zephyr-eabi-gdb: error while loading shared libraries: libpython3.6m.so.1.0: cannot open shared object file: No such file or directory

this is because on recent debian/ubuntu 3.6 headers don't seem to be available.  See:

https://tracker.debian.org/pkg/python3.6

there are some "workarounds"  out there .. like:

https://stackoverflow.com/questions/43621584/why-cant-i-install-python3-6-dev-on-ubuntu16-04

but at least for me I still can't get the 3.6 headers (3.7 headers ARE easily installed with package python3-dev ...)

stephanosio:
The GDB supports Python scripting through the use of libpython, which
is dynamically linked.

Since there are multiple widely used Python 3.x versions and the
vanilla package support for them varies across Linux distros, force
linking against the libpython2.7, which practically is universally
supported in all modern Linux distros.

Fixes #165
Fixes #196


galak:
Created this PR #203 to try static gdb + python.  Lets see what we get.

nashif:
>  force
> linking against the libpython2.7, which practically is universally
> supported in all modern Linux distros.

Is this true though? Python 2 is starting to be deprecated on almost all operating systems and distros. Why are we going back to 2.7?

stephanosio:
> Python 2 is starting to be deprecated on almost all operating systems and distros.

@nashif Sure they are deprecated, but that does not mean they will stop providing packages for it. I checked Ubuntu 20.04 LTS, Debian 10 and CentOS 8, and they are still providing the Python 2.7 package.

As long as we are dynamic linking Python that is the most viable option. @galak is looking into static linking, so if that works, we should be able to get Python 3.x in.

tejlmand:
> > Python 2 is starting to be deprecated on almost all operating systems and distros.
> 
> @nashif Sure Python 2.7 is deprecated, but that does not mean they will stop providing packages for it. I checked Ubuntu 20.04 LTS, Debian 10 and CentOS 8, and they are still providing the Python 2.7 package.

I agree with @nashif here.
I think it is a bad idea to go back to Python 2.7.
It was sunset on 1st Jan 2020.
> We have decided that January 1, 2020, was the day that we sunset Python 2. That means that we will not improve it anymore after that day, even if someone finds a security problem in it. You should upgrade to Python 3 as soon as you can.

https://www.python.org/doc/sunset-python-2/
Not even security issues will be fixed.

So linking to Python 2.7 is just asking for troubles in the new future.

stephanosio:
well, given that making GDB use Python 3.6 was not a deliberate/conscious decision, but rather an unintentional side effect of migrating to a Ubuntu 18.04-based CI image, I do not necessarily think reverting GDB to use Python 2.7 **until we have a more permanent solution** (e.g. static linking or `libpython3.so`) is a bad thing (that is why I suggested this for `0.11.3` rather than `0.12.0`).

The status quo is that many users are forced to install third-party (i.e. not from the distro) packages to make GDB work because a distro-provided Python 3.6 package is not available.

Also please note that this does not affect the rest of SDK, it only affects the GDB Python scripting capability.

carlescufi:
@galak and @stephanosio if  we need to pick a Python version to link against, I suggest we instead pick Python 3.7, which will be stable for a while. Using Python 2 forces users to maintain two sets of pip packages, since all zephyr dependencies use Python 3.

stephanosio:
@carlescufi Not sure if you read the issues related to this PR, but the reason for forcing Python 2.7 instead of 3.x is that vanilla package support for Python 3.x varies across distros/versions and the Zephyr SDK users are currently forced to install third-party Python packages (or compile their own).
image
_(N/A means no distro-provided package is available)_

This, in theory, can be solved by linking against `libpython3.so` instead of `libpython3.x.so`, but this version-independent library is unfortunately not available in Ubuntu and other common distros.

Another way to solve this problem is to include a specific version of Python 3 in the Zephyr SDK, but due to the limitations of the crosstool-ng and our SDK build system, this is easier said than done -- the toolchains (including `gdb`) are currently built by crosstool-ng with the host sysroot, but we need to build Python 3 in Yocto and force the Poky native SDK sysroot; after that, we need to either relative-RPATH the Poky `libpython3.x.so` (not really an option if we have glibc and/or other lib version incompatibility), or set Poky-specific dynamic loader path in the GDB program header (must be fully self-contained).

There is also an additional problem -- that the self-contained nature of the Poky native SDK library system is currently broken and we will need to fix that before we can provide Zephyr SDK-specific `libpython3.x.so` that can run on the Poky ecosystem.

carlescufi:
@stephanosio thanks for the extensive explanation on the different support for Python version in multiple distributions. The fact that we cannot link against `libpython3.so` because major distros do not include it is indeed a problem. But based on the information on your table, I believe linking against Python 3.8.x (i.e. `libpython3.8.so`) is the right move right now because:

- It gives us support for Ubuntu releases going back 2 years in time
- Debian testing tends to be used by desktop users, not stable nor unstable, and that has support for [Python 3.8](https://packages.debian.org/testing/python/python3)

As a user, I'd rather be forced to update my Python version to maintained, secure one, than having to live with an unsecure Python 2.7. The other issue is that some GDB Python frameworks might drop support for Python 2.7 in the near future, complicating the issue even further.

galak:
> @stephanosio thanks for the extensive explanation on the different support for Python version in multiple distributions. The fact that we cannot link against `libpython3.so` because major distros do not include it is indeed a problem. But based on the information on your table, I believe linking against Python 3.8.x (i.e. `libpython3.8.so`) is the right move right now because:
> 
> * It gives us support for Ubuntu releases going back 2 years in time
> * Debian testing tends to be used by desktop users, not stable nor unstable, and that has support for [Python 3.8](https://packages.debian.org/testing/python/python3)
> 
> As a user, I'd rather be forced to update my Python version to maintained, secure one, than having to live with an unsecure Python 2.7. The other issue is that some GDB Python frameworks might drop support for Python 2.7 in the near future, complicating the issue even further.

I'm curious what python versions look like on Fedora and Arch as those seem to be 2 other distros people have used.

galak:
Trying python3.8 over at #210

galak:
Closing this in favor of building against python3.8 and building a no-py version of gdb.

galak:
Closing this as we decided to use python3.8 to try and get support on the most systems with python3.x

PR #204, opened 2020-04-17
qemu: Add support for changing CPU type in sifive_e machine.

xodus7:
This change allows the sifive_e machine to run with different CPU
targets to enable different ISA entensions. To that end it also
introduces a new sifive-e34 CPU type which provides the same ISA
as sifive-e31, with the addition of the single precision
floating-point extension (f). The default CPU for the sifive_e
machine is unchanged.

Signed-off-by: Corey Wharton <coreyw7@fb.com>

### DIFF
--- a/meta-zephyr-sdk/recipes-devtools/qemu/files/0005-riscv-sifive_e-Support-changing-CPU-type.patch
+++ b/meta-zephyr-sdk/recipes-devtools/qemu/files/0005-riscv-sifive_e-Support-changing-CPU-type.patch
@@ -1,0 +1,39 @@
+From 6bee5cdeb7ae297a997e23ab671beaaf3b3ebf58 Mon Sep 17 00:00:00 2001
+From: Corey Wharton <coreyw7@fb.com>
+Date: Thu, 12 Mar 2020 15:38:16 -0700
+Subject: [PATCH v2 1/2] riscv: sifive_e: Support changing CPU type
+
+Allows the CPU to be changed from the default via the -cpu command
+line option.
+
+Signed-off-by: Corey Wharton <coreyw7@fb.com>
+Reviewed-by: Bin Meng <bmeng.cn@gmail.com>
+Reviewed-by: Alistair Francis <alistair.francis@wdc.com>
+---
+ hw/riscv/sifive_e.c | 3 ++-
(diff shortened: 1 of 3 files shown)


galak:
Sweet, thanks for doing this!

galak:
@xodus7 do you know if we can configure the "DTIM" size via the command line? 

(would be very useful to increase it beyond the 16k limit we currently have)

xodus7:
@galak It's hard-coded in the machine memory map:

```c
static const struct MemmapEntry {
    hwaddr base;
    hwaddr size;
} sifive_e_memmap[] = {
    // ....
    [SIFIVE_E_DTIM] =     { 0x80000000,     0x4000 }
};
```

galak:
> @galak It's hard-coded in the machine memory map:
> 
> ```c
> static const struct MemmapEntry {
>     hwaddr base;
>     hwaddr size;
> } sifive_e_memmap[] = {
>     // ....
>     [SIFIVE_E_DTIM] =     { 0x80000000,     0x4000 }
> };
> ```

Twist your arm to work up a patch to make it configurable?

xodus7:
> Twist your arm to work up a patch to make it configurable?

I'm not familiar enough with QEMU to know how much work that would be and don't have time in the near future to look into it. Would it be acceptable to expand the DTIM in the map for our purposes? We'd only then need to change the dtim node in the board device tree file.

galak:
> I'm not familiar enough with QEMU to know how much work that would be and don't have time in the near future to look into it. Would it be acceptable to expand the DTIM in the map for our purposes? We'd only then need to change the dtim node in the board device tree file.

No worries, updating only the DTS isn't sufficient since we need the memory to exist when we run qemu.

xodus7:
@galak I meant change the memory map in the machine _and_ update the device tree. Basically rather than make it a command line option just patch the machine to 64-128kB directly.

galak:
closed/re-opened PR to get CI to run again.

carlescufi:
@galak is this ready to merge?

galak:
> @galak is this ready to merge?

CI was semi-broken for a while.  Need to test this before merging, but the change lgtm.

xodus7:
FYI, these patches were pulled into the QEMU RISC-V tree for v5.1.

PR #206, opened 2020-04-23
Fix use of CRC32 intrinsics with Armv8-a and hard-float

galak:
Backport patch from gcc to fix issue with building w/arm_acle.h.


MaureenHelm:
Confirmed this fixes the zephyr sdk issue in zephyrproject-rtos/zephyr#24655. Thanks @galak

PR #212, opened 2020-05-12
template_dir: improved help text for toolchain install dir

tejlmand:
This commit reduces number of recommended folders when user is
installing outside any of the default paths.

For example, is user uses a path under /usr/ but not a default search
path, such as /usr/mysdk, then only the /usr/ based recommended paths
are printed.

Also a link to Zephyr docs for all search paths are given.

Signed-off-by: Torsten Rasmussen <Torsten.Rasmussen@nordicsemi.no>

### DIFF
--- a/scripts/template_dir
+++ b/scripts/template_dir
@@ -246,14 +246,25 @@ if [    "$sdk_dirname" != "/opt" \
 	-a "$sdk_dirname" != "$HOME/bin" ] ||\
 	[[ "${sdk_basename}" != "zephyr-sdk"* ]]; then
 	echo
-	echo "It is recommended to install Zephyr SDK at one of the following locations:"
-	echo "  ${HOME}/zephyr-sdk[-${SDK_VERSION}]"
-	echo "  ${HOME}/.local/zephyr-sdk[-${SDK_VERSION}]"
-	echo "  ${HOME}/.local/opt/zephyr-sdk[-${SDK_VERSION}]"
-	echo "  ${HOME}/bin/zephyr-sdk[-${SDK_VERSION}]"
-	echo "  /opt/zephyr-sdk[-${SDK_VERSION}]"
-	echo "  /usr/zephyr-sdk[-${SDK_VERSION}]"
-	echo "  /usr/local/zephyr-sdk[-${SDK_VERSION}]"
+	echo "It is recommended to install Zephyr SDK at one of the following locations for automatic discoverability in CMake:"
+
+	if [ ${sdk_dirname:0:4} == "/usr" ]; then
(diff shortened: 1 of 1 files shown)


tejlmand:
@nashif @stephanosio As suggested on slack, this simplifies the help text, and makes it easier to copy paste.
Per default, it will keep the `SDK version` in the printed path, as it is easier for a user to NOT copy version, than to add it.

It also reduces the wall of text, as it only prints recommended folders, based on where user is already considering to install.
If user has set install folder to: `/usr/my-sdk/` then it will only print:
```
It is recommended to install Zephyr SDK at one of the following locations for
automatic discoverability in CMake:
  /usr/zephyr-sdk-0.11.3
  /usr/local/zephyr-sdk-0.11.3

Note: The version number '-0.11.3' can be omitted.

For full list of recommended folders, please go to:
https://docs.zephyrproject.org/latest/getting_started/installation_linux.html
```

but if start of path was under `~/my-sdk`, it will print the search paths under `~`, as:
```
It is recommended to install Zephyr SDK at one of the following locations for
automatic discoverability in CMake:
  /home/tora/zephyr-sdk-0.11.3
  /home/tora/bin/zephyr-sdk-0.11.3

... (same as above message ^^^)
```

Default, if nothing is matched, such as installing to `/tmp` will be:
```
It is recommended to install Zephyr SDK at one of the following locations for
automatic discoverability in CMake:
  /home/tora/zephyr-sdk-0.11.3
  /opt/zephyr-sdk-0.11.3

... (same as above message ^^^)
```

tejlmand:
i'm unsure about adding this:
```
For full list of recommended folders, please go to:
https://docs.zephyrproject.org/latest/getting_started/installation_linux.html
```
so specific feedback on this is very welcome.

galak:
> i'm unsure about adding this:
> 
> ```
> For full list of recommended folders, please go to:
> https://docs.zephyrproject.org/latest/getting_started/installation_linux.html
> ```
> 
> so specific feedback on this is very welcome.

I'd drop this, worry about it being out of date or point to the wrong thing in the future.

tejlmand:
> 
> 
> > i'm unsure about adding this:
> > ```
> > For full list of recommended folders, please go to:
> > https://docs.zephyrproject.org/latest/getting_started/installation_linux.html
> > ```
> > 
> > 
> > so specific feedback on this is very welcome.
> 
> I'd drop this, worry about it being out of date or point to the wrong thing in the future.

Exactly my concern as well.
Will drop.
Just feel it's would be nice to inform there are more options, without the previous too long list.

tejlmand:
> I'd drop this, worry about it being out of date or point to the wrong thing in the future.

done

PR #214, opened 2020-05-13
Build GDB w/ & w/o python support

galak:
The gdb w/o python is called `{PREFIX}-gdb-no-py`


nashif:
> The gdb w/o python is called `{PREFIX}-gdb-no-py`

why not the other way, ie. gdb that always works and what most tools expect, and gdb-py if someone wants python?

galak:
> > The gdb w/o python is called `{PREFIX}-gdb-no-py`
> 
> why not the other way, ie. gdb that always works and what most tools expect, and gdb-py if someone wants python?

I didn't want to break people's expectations of `{PREFIX}-gdb` works for a "minor" release.  I can document that the name will change in the next major release to be `gdb` and `gdb-py`.

PR #216, opened 2020-05-15
Build arm ghub

galak:


galak:
this is completely grabbing the github action work that @stephanosio did here https://github.com/stephanosio/zephyr-crosstool-ng/blob/master/.github/workflows/ci.yml.  The is purely to see how long this takes on GitHub CI vs shippable.

PR #223, opened 2020-05-22
newlib: Make configure host consistent for all arch's

galak:
Add a zephyr case to newlib/configure.host so we get the same config on
all arch's.  Fix the fact that some arch's like x86 were getting
-DMISSING_SYSCALL_NAMES set in newlib which we don't want.

Fixes: #221

### DIFF
--- a/patches/newlib/3.1.0.20181231/0004-newlib-configure.host-Add-host-config-for-zephyr.patch
+++ b/patches/newlib/3.1.0.20181231/0004-newlib-configure.host-Add-host-config-for-zephyr.patch
@@ -1,0 +1,40 @@
+From 2599d2e1016beeb8dcfe691940acc334bdc6eb24 Mon Sep 17 00:00:00 2001
+From: Kumar Gala <kumar.gala@linaro.org>
+Date: Thu, 21 May 2020 22:55:16 -0500
+Subject: [PATCH] newlib/configure.host: Add host config for zephyr
+
+Add zephyr specific host config so that all arch toolchain variants we
+build for zephyr behave the same way.  (otherwise we end up getting
+-DMISSING_SYSCALL_NAMES set on some arch's like NIOS and x86, but not on
+others).
+
+Signed-off-by: Kumar Gala <kumar.gala@linaro.org>
+---
+ newlib/configure.host | 11 +++++++++++
+ 1 file changed, 11 insertions(+)
+
+diff --git a/newlib/configure.host b/newlib/configure.host
+index 37d2b9b2e..77e855e85 100644
+--- a/newlib/configure.host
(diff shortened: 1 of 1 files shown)


stephanosio:
@galak This is for 0.11.4, I suppose?

galak:
> @galak This is for 0.11.4, I suppose?

Yeah, that was my thinking.

patches/newlib/3.1.0.20181231/0004-newlib-configure.host-Add-host-config-for-zephyr.patch
```diff
@@ -0,0 +1,40 @@
...
+@@ -665,6 +665,17 @@ newlib_cflags="${newlib_cflags} -DCLOCK_PROVIDED -DMALLOC_PROVIDED -DEXIT_PROVID
+   *-*-tirtos*)
+ 	newlib_cflags="${newlib_cflags} -D__DYNAMIC_REENT__ -DMALLOC_PROVIDED"
+ 	;;
++  *-zephyr-*)
++	syscall_dir=syscalls
++	case "${host}" in
++	  arm*-*-*)
++		if [ "x${newlib_may_supply_syscalls}" = "xyes" ] ; then
++		  newlib_cflags="${newlib_cflags} -DARM_RDI_MONITOR"
```
pfalcon, line 29: @galak, Is this the same patch which was here originally? I briefly looked at it when it was posted, but now looking in detail, I'm not sure I understand what happens here. The patch description talks about being consistent with passing (or not?) -DMISSING_SYSCALL_NAMES for all platforms, but the patch content actually passes -DARM_RDI_MONITOR for `arm*-*-*` platforms only. What's -DARM_RDI_MONITOR and how it's related to the original issue of syscall() vs _syscall() naming?

pfalcon: > Add zephyr specific host config

Ah, seems I'm starting to understand now. Previously, there was no config case for `*-zephyr-*` specifically, so there was random matchings based on arch patterns as sprinkled around in upstream. Now we have common `*-zephyr-*` switch case (and thus consistent config options), but need to pass -DARM_RDI_MONITOR  specifically for ARM platforms.

pfalcon: So, this is definitely useful, we just need to understand, whether it's better to have -DMISSING_SYSCALL_NAMES  or -UMISSING_SYSCALL_NAMES ;-).

pfalcon:
@stephanosio: Hopefully you can approve too.

PR #231, opened 2020-06-23
x86_64: newlib: use march=pentium for 32-bit

dcpleung:
The 32-bit build on Zephyr main tree is targeting Pentium-based
devices. However, given the target tuple, newlib is being built
targeting i686 by default. This emits instructions that cannot
run on Pentium-based devices. So add a march=pentium when
newlib is being built for x86_64 32-bit.

### DIFF
--- a/patches/newlib/3.1.0.20181231/0003-Support-building-for-32-bit-under-x86_64.patch
+++ b/patches/newlib/3.1.0.20181231/0003-Support-building-for-32-bit-under-x86_64.patch
@@ -1,4 +1,4 @@
-From dc48595340a292f044a0fddcab2d67a81650010f Mon Sep 17 00:00:00 2001
+From 7fa9f8eb8aff8bd73d04dd04f72eeac4682e38dc Mon Sep 17 00:00:00 2001
 From: Daniel Leung <daniel.leung@intel.com>
 Date: Wed, 8 May 2019 09:17:10 -0700
 Subject: [PATCH 3/3] Support building for 32-bit under x86_64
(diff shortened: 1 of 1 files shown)


dcpleung:
> Are there any Zephyr-compatible x86 devices that do not support the MMX instruction set?

Not that I know of. Even the old Arduino 101 has MMX support.

stephanosio:
> Not that I know of. Even the old Arduino 101 has MMX support.

In that case, I wonder if it would be feasible to make the Zephyr `-march` type `i686` (i.e. Pentium Pro) instead:
https://github.com/zephyrproject-rtos/zephyr/blob/b7eb04b3007767be9febe677924918d2422a4406/soc/x86/ia32/CMakeLists.txt#L9

That is, of course, unless it is foreseeable that there will be a future x86 SoC without MMX support.

stephanosio:
For reference:

Besides MMX,
image

dcpleung:
Pentium has MMX support. However, there are SoCs which do not support Pentium Pro's ISA. It is why newlib needs to be compatible with Pentium but not Pentium Pro. CMOV* are the ones not supported in Pentium.

stephanosio:
> Pentium has MMX support. However, there are SoCs which do not support Pentium Pro's ISA. It is why newlib needs to be compatible with Pentium but not Pentium Pro. CMOV* are the ones not supported in Pentium.

So, IIUC, MMX is available on all compatible SoCs but CMOV* aren't.

In that case, should it be using `-march=pentium-mmx` instead of `-march=pentium` so as to allow utilisation of the MMX instruction set while keeping things i586/P5?

dcpleung:
> > Pentium has MMX support. However, there are SoCs which do not support Pentium Pro's ISA. It is why newlib needs to be compatible with Pentium but not Pentium Pro. CMOV* are the ones not supported in Pentium.
> 
> So, IIUC, MMX is available on all compatible SoCs but CMOV* aren't.
> 
> In that case, should it be using `-march=pentium-mmx` instead of `-march=pentium` so as to allow utilisation of the MMX instruction set while keeping things i586/P5?

I always thought "-march=pentium" had MMX support. Let me test with "pentium-mmx" and make sure it works.

stephanosio:
AFAIK, the discontinued Quark (Lakemont) was the only P5-based SoC (hence, no MMX/CMOV support).

Are there any other P5-based SoCs? (unless you are referring to the "Pentium" in the UP Squared which is not P5-based).

dcpleung:
There are Lakemont based SoCs (like the ones in the Integrated Sensor Hub). So I think we should stick with "pentium" for now.

stephanosio:
> There are Lakemont based SoCs (like the ones in the Integrated Sensor Hub). So I think we should stick with "pentium" for now.

Fair enough, it seems we would eventually need to enable multilib support for the `x86_64-zephyr-elf` target to properly address this.

PR #252, opened 2020-07-18
scripts/template_dir (setup.sh): fixes for unattended setup for distro packagers

alexeicolin:
Fixes a few issues encountered when scripting the installation for distro package:
1. Issue introduced when cmake module was added: the user was prompted even if -y was passed, which broke unattended installation.
2. CMake module installed under $HOME unconditionally, and the path to SDK in the cmake module equals destination directory, which is not necessarily the same as the final installation path in the system (details below). This issue is not solved, but only worked around by adding -[no]cmake args so that the problematic cmake module generation can be disabled and performed out-of-band by the packager script. 
3. Order of -y and -[no]rc  arguments mattered due to -y unintentionally overriding -[no]rc. Overlooked in 6dd3e26c271bbf139f6d30523bc55fc7673b52f5.
4. Duplication of code for prompting user

Tested these patches applied on top of 0.11.4 release, did not test in master.

Suggesting to review one commit at a time and also suggest not to squash the commits when merging:
 
```
commit 706b10cddfd781e3c7f1006a64ad23e09601e77c
    scripts/template_dir: -[no]cmake flag for CMake module installation
    
    There are two issues with CMake module installation that are
    problematic for distro packages:
    1. the cmake module is installed under $HOME and this path is not overridable
    2. the path to SDK recorded in the cmake file is the same as the
    target directory, which might be prefixed by the staging path of the
    package build (e.g.  ~/mypackages/zerphyr-sdk/destdir/opt/zerphyr-sdk),
    not the final installation directory in the system (/opt/zephyr-sdk).
    
    1 is fixable by making the variable overridable from the environment:
        : ${ZEPHYR_SDK_REGISTRY_DIR:=$HOME/.cmake/packages/Zephyr-sdk}
    (but this may be a bashism). But, 1 is not useful unless 2 is also
    addressed.
    
    2 might be fixable by introducing a prefix argument -p (default "") in
    addition to the target_dir (-d), and composing the destination directory
    for copying files as $prefix/$target_dir while using only $target_dir
    wherever installation paths are written, such as in the CMake module.
    
    Until 1 or 2 are addressed, add -[no]cmake arg so that the package
    scripts can install the CMake module themselves.

commit 03933375277558f939a12990cae3ee372511d87e
    scripts/template_dir: re-use common prompting code

commit e2e27bb5c5c044203c6085279a99514416799c51
    scripts/template_dir: don't prompt for cmake when -y set

commit b8d47d156f3e89237a80b349eb2ec48160046b05
    scripts/template_dir: don't let -y override -[no]rc
```

### DIFF
--- a/scripts/template_dir
+++ b/scripts/template_dir
@@ -66,7 +66,9 @@ while [ "$1" != "" ]; do
 			;;
 		-y )
 			confirm="y";
-			rc_confirm="y";
+			if [ -z "$rc_confirm" ]; then
+			  rc_confirm="y"
+			fi
 			;;
 		-rc )
 			rc_confirm="y";
--- a/scripts/template_dir
+++ b/scripts/template_dir
@@ -144,7 +144,7 @@ do_zephyrrc()
 	echo "     export ZEPHYR_SDK_INSTALL_DIR=$target_sdk_dir"
 	echo
 	if [ -z "$rc_confirm" ]; then
-		prompt "creating/updating $HOME/.zephyrrc with environment variables setup for you "
+		prompt "create/update $HOME/.zephyrrc with environment variables setup for you "
 		rc_confirm=$answer
 	fi
 	if [ "$rc_confirm" = "y" ]; then
@@ -172,7 +172,7 @@ do_cmake_package()
 		fi
 
 		if [ -z "$cmake_confirm" ]; then
-			prompt "registering Zephyr-sdk CMake module with path $target_sdk_dir "
+			prompt "register Zephyr-sdk CMake module with path $target_sdk_dir "
 			cmake_confirm=$answer
 		fi
 		if [ "$cmake_confirm" = "y" ]; then
@@ -188,7 +188,7 @@ do_cmake_package()
 # Read the input "y"
 # $1 is optional, but allows the caller to provide additional text.
 prompt () {
-	echo "Do you want to continue $1(y/n)? "
+	echo "Do you want to $1(y/n)? "
 	while read answer; do
 		if [ "$answer" = "Y" -o "$answer" = "y" ]; then
 			answer="y"
--- a/scripts/template_dir
+++ b/scripts/template_dir
@@ -163,15 +163,7 @@ do_cmake_package()
 			return
 		fi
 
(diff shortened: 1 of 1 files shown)


scripts/template_dir
```diff
@@ -182,17 +187,28 @@ do_cmake_package()
...
-      while read confirm; do
-          [ "$confirm" = "Y" -o "$confirm" = "y" -o "$confirm" = "n" \
-            -o "$confirm" = "N" ] && break
-          echo "Invalid input \"$confirm\", please input 'y' or 'n': "
-      done
-  else
-      echo
-  fi
+prompt () {
+	echo "Do you want to continue $1(y/n)? "
```
yperess, line 191: The wording here is a bit confusing, if I saw:

"Do you want to continue registering Zephyr-sdk CMake module with path... (y/n)?"

I would have thought that hitting 'n' would have terminated the setup, Can we leave out the prefix and only standardize the "(y/n)?"

yperess (changes requested):
This is great, I'd really like to see this merge as it would help out being able to install the sdk without interaction (e.g. in an ebuild)

scripts/template_dir
jackrosenthal: +1 on this

jackrosenthal (changes requested):
Good to see this.

scripts/template_dir
galak: @alexeicolin any update on this feedback?

alexeicolin: Haven't got to this yet, I'll take a look.

galak: thanks, would like to get this in for 0.12.0, so not urgent, but would like to get this merged in the next week or 2 if possible.

sjg20 (approved):

scripts/template_dir
alexeicolin: Removed the word 'continue' to prevent user confusion (pushed this as separate commit, because didn't really fit with any of the other commits).

alexeicolin:
Addressed feedback and rebased. (Did not retest after rebase.)

yperess (approved):

PR #257, opened 2020-07-21
xtensa: newlib: fix PRI.PTR macros for correct intptr_t type

dcpleung:
The PRI.PTR macros are using int instead of long int for
intptr_t (and also uintptr_t). This has been fixed upstream
so pulling in the necessary bits from 3.1.


galak:
Should be handled by update to newlib 3.3 in pr #258

PR #263, opened 2020-09-03
qemu: Add MPS2_AN386 support

galak:
Add support for the AN386 which is a Cortex-M4 based board.


galak:
Closing this as the hope was this would get us floating point support, but QEMU doesnt support floating point for Cortex-M yet.

PR #271, opened 2020-09-19
Add picolibc config files

keith-packard:
These configs direct crosstool-ng to use picolibc instead of newlib or newlib-nano. Note that there is a bug in upstream crosstool-ng which prevents the riscv64 or arm64 variants from working, a suitable PR has already been merged to fix those. However, the arm configuration works just fine with the current crosstool-ng version referenced by sdk-ng, and with the fix applied, the riscv and aarch64 toolchains also work.


galak:
> These configs direct crosstool-ng to use picolibc instead of newlib or newlib-nano. Note that there is a bug in upstream crosstool-ng which prevents the riscv64 or arm64 variants from working, a suitable PR has already been merged to fix those. However, the arm configuration works just fine with the current crosstool-ng version referenced by sdk-ng, and with the fix applied, the riscv and aarch64 toolchains also work.

Got a reference to the upstream fix PR?

keith-packard:
> Got a reference to the upstream fix PR?

Sure, I should have linked that already.

https://github.com/crosstool-ng/crosstool-ng/pull/1390

galak:
PR #272 will pull in updated crosstool-ng w/the picolibc updates.

Do we think its feasible to have picolibc built along side newlib and newlib_nano so we don't need yet another toolchain package?

keith-packard:
> Do we think its feasible to have picolibc built along side newlib and newlib_nano so we don't need yet another toolchain package?

This isn't supported by crosstool-ng directly, but picolibc can be co-installed alongside newlib in the same tree. Upstream work in crosstool-ng to support multiple C libraries or downstream work in sdk-ng to add picolibc to an existing crosstools-ng tree would both be feasible.

adfernandes:
@keith-packard Thank you for doing this! Just this past weekend I was thinking "Wow, it would be really nice to be able to use `picolibc` with Zephyr (and this toolchain)..." :+1: :smile:

keith-packard:
> @keith-packard Thank you for doing this! Just this past weekend I was thinking "Wow, it would be really nice to be able to use `picolibc` with Zephyr (and this toolchain)..." +1 smile

You're welcome! I'm busy testing this with the latest picolibc release (1.4.7) which has a couple of bug fixes that should make the zephyr tests pass at last! The biggest blocker here is that it would be nice to be able to build all three libraries with sdk-ng so that the test system could check all three with the same toolchain.

galak:
> > 
> 
> You're welcome! I'm busy testing this with the latest picolibc release (1.4.7) which has a couple of bug fixes that should make the zephyr tests pass at last! The biggest blocker here is that it would be nice to be able to build all three libraries with sdk-ng so that the test system could check all three with the same toolchain.

Hopefully we can make some progress on https://github.com/crosstool-ng/crosstool-ng/issues/1395.

keith-packard:
I've managed to get crosstool-ng to include picolibc with existing builds; I've got a new PR for sdk-ng using that, so I'll close this PR.

PR #287, opened 2020-11-02
Add picolibc to existing newlib/newlib-nano configurations

keith-packard:
This takes advantage of a PR I posted to the zephyr crosstool-ng project which changes picolibc into a companion library, allowing it to be added to any existing build without changing anything else.


d3zd3z:
Not exactly a direct comment on this PR, but I wanted to bring up another question. Zephyr depends on newlib (or possibly picolibc). Right now we build newlib as part of our SDK. This makes finding and fixing issues caused by the C library fairly challenging, even though we depend on quite a bit of our posix behavior coming from this library.

What do people think of the idea of putting the C library as a module that gets built as part of Zephyr builds.

keith-packard:
Does zephyr support submodules? Picolibc currently builds using meson; would it be useful to add cmake support to build for a single target be useful for this?

stephanosio:
> Does zephyr support submodules? Picolibc currently builds using meson; would it be useful to add cmake support to build for a single target be useful for this?

Not "submodules" as in the git sense, but we support external modules through something called "west". See https://github.com/zephyrproject-rtos/zephyr/blob/main/west.yml.

> What do people think of the idea of putting the C library as a module that gets built as part of Zephyr builds.

As briefly noted in the chat, for newlib which is the default C library for the Zephyr SDK, it is not really feasible for multiple reasons described there.

As for the picolibc though, I think it might be of use to add it as a module so we can have a more fully featured toolchain independent C library besides the "minimal libc" we currently have in Zephyr (of course, the "toolchain independent" part would need to be further investigated).

keith-packard:
picolibc builds with three compilers at present: gcc, clang and compcert's compiler, and is tested with gcc and clang. Adding cmake build infra to work as a git submodule within Zephyr should be straightforward, and could be upstreamed and tested in picolibc's usual testing framework.

keith-packard:
> Not "submodules" as in the git sense, but we support external modules through something called "west".

I didn't know west could do that! Let me know how I could help make this work.

d3zd3z:
How we bring in various modules depends a bit on the build system of that module. In some cases, we use their CMake files. In other cases (such as Mbed TLS, currently) we have our own CMake files that describe that build.

One concern about doing this that was raised on discord was concerning the increase in build time due to having to build the C library. It is an argument for continuing to provide the C library through the SDK.

keith-packard:
> How we bring in various modules depends a bit on the build system of that module. In some cases, we use their CMake files. In other cases (such as Mbed TLS, currently) we have our own CMake files that describe that build.

I'm happy to integrate CMake files into picolibc, although just using meson would be easiest for me.

> One concern about doing this that was raised on discord was concerning the increase in build time due to having to build the C library. It is an argument for continuing to provide the C library through the SDK.

Picolibc builds for a single architecture take only a few seconds, including configuration. Prob ably worth it just to reduce the complexity of the SDK.

keith-packard:
This change is still useful as crosstool-ng has the infrastructure necessary to build libstdc++ as well as libc, so I think it's reasonable to include picolibc in the SDK, especially now that crosstool-ng has picolibc support available.

stephanosio (commented):
I am in favour of including the picolibc as a Zephyr module with the goal of ultimately replacing the Zephyr's current minimal libc.

See https://github.com/picolibc/picolibc/issues/177#issuecomment-1073815655.

keith-packard:
As discussed in discord, this PR is still useful in providing C++ support using picolibc as that requires the toolchain to provide libstdc++ that works with a specific build of the C library.

evgeniy-paltsev:
Hi @keith-packard ,

just noted that you add this improvement for `configs/arc-zephyr-elf.config` but not for `configs/arc64-zephyr-elf.config`. I'm wondering if there is any specific reason for that?

Or it's just WIP and I'm checking it too soon? :)

keith-packard:
> Hi @keith-packard ,
> 
> just noted that you add this improvement for `configs/arc-zephyr-elf.config` but not for `configs/arc64-zephyr-elf.config`. I'm wondering if there is any specific reason for that?
> 
> Or it's just WIP and I'm checking it too soon? :)

I haven't tried arc64 yet; have you?

stephanosio:
The picolibc configs are being (silently) disabled because the build machine is missing ninja and meson:

https://github.com/crosstool-ng/crosstool-ng/blob/29b6e00368b4f74c4d4a1cd22da7fdabdcd1d6ea/config/comp_libs/picolibc.in#L3

```
checking for meson... no
checking for ninja... no
```

ct-ng should really warn about selected, but set to `n` configs ....

stephanosio:
Rebased

keith-packard:
ok, I've tested the most recent build and it now actually works. Was having troubles with include/library paths in picolibc.specs that I've sorted out. I *think* this PR can be considered for merging, if we can figure out what it should look like.

keith-packard:
I've removed the DNM label, not because this *should* be merged, but because something much like it *could* be merged and it would work.

keith-packard:
With https://github.com/zephyrproject-rtos/zephyr/issues/49922 blocked on this PR, it's time to figure out what the SDK should look like with picolibc included. Do we

 a. Just add picolibc to the builds and accept the disk space increase?
 b. Split out some libc subsets to separate packages that can be downloaded independently?
 c. Build separate downloads for each architecture?

Separate, but somewhat connected from this question is what should be done with debug info -- right now, the SDK removes all of the debugging symbols, which makes diagnosing bugs that cross the libc boundary much more difficult.  However, including them in the SDK would consume a lot of disk space. I suspect that simply including them in an all-arch all-libc bundle would end up being way too big. Just splitting across C libraries would probably not reduce the size enough. However, I think that splitting across architectures probably would be sufficient to keep each one under 2GB. 

The alternative would be to create separate debug packages that can be downloaded independently. Even separated from the binaries, I suspect these would be large if they included all archtectures and all libc versions. So we might need to split these, either per-libc or per-arch.

My recommendation would be to split the SDK into per-architecture packages and include debug symbols directly in those packages. Most developers use only a few architectures, a complete SDK for other architectures would be much less likely to have useful bits than including C library debug symbols.

Yes, this would make the combined bits needed for testing significantly larger; one obvious option would be to create an all-arch all-libc no-debug package that can be used for CI.

stephanosio:
@keith-packard it looks like there is an issue with Canadian Cross compilation.

keith-packard:
> @keith-packard it looks like there is an issue with Canadian Cross compilation (note that aarch64 host toolchains are built from x86-64 host).

Yup, just looks like the --prefix argument isn't getting computed correctly in this case. Happy to debug this locally, can we extract the crosstool-ng configuration file for this piece somehow?

stephanosio:
re: https://github.com/zephyrproject-rtos/sdk-ng/pull/287#issuecomment-1238359430

> b. Split out some libc subsets to separate packages that can be downloaded independently?

That is one option, but it adds more complexity in terms of maintenance and usage. It can also be potentially confusing to most users (e.g. people who are new to Zephyr may not fully not understand what newlib and picolibc are and the ramifications of using them).

> c. Build separate downloads for each architecture?

In my opinion, this is the most straight-forward and clean approach; in fact, we already do that to some degree (note that individual toolchains are downloadable from the release assets). They are not meant to be directly downloaded and used by the end users (though, they _can_ if they wanted to and know what to do with it) -- they are downloaded by the "minimal" distribution bundle setup script, which allows the user to select the toolchains to download and install.

> a. Just add picolibc to the builds and accept the disk space increase?

The main issue with that is that GitHub does not allow uploading release assets larger than 2GB and we will not be able to upload the "all-in-one" distribution bundles to the GitHub releases with the picolibc also bundled in the toolchains.

We can resolve this in one of the following ways:

1. Upload the "all-in-one" distribution bundles to a different file hosting service (e.g. S3).
    - While this would be the simplest solution, it will incur a lot of extra cost in terms of outbound traffic.
    - With the size of the distribution bundles and the number of downloads, this can be in the range of thousands of dollars per month.
1. Use a better compression algorithm
    - Zephyr SDK currently uses GZip (tar.gz) for Linux and macOS, and DEFLATE (zip) for Windows.
    - GZip and DEFLATE are not exactly known for having the best data compression ratio.
    - Using LZMA should give much better compression ratio (a quick test has shown around 1.5x for the "all-in-one" distribution bundle with xz-utils).
    - Use XZ (tar.xz) for Linux and macOS, and 7z/LZMA (7z) for Windows. This should keep the "all-in-one" distribution bundle sizes to below 2GB.
    - LZMA requires more processing power and takes longer to compress and decompress, but this is not so much as to prevent its adoption (a quick test has shown that decompressing "all-in-one" distribution bundle tar.gz took approx. 1 minute whereas tar.xz took approx. 2.5 minutes on my machine, which is not that bad).
    - Of course, even after switching to LZMA, future changes may increase the distribution bundle sizes to more than 2GB, in which case we will have to evaluate the multi-volume archive approach described below.
1. Use multi-volume archives
    - As its name states. Split single compressed archive into multiple volumes/parts, which can be combined together after download and decompressed.

keith-packard:
Even though the build failed when it ran out of space doing the Windows testing, I was able to extract the canadian cross built aarch64 linux toolchain for arm targets and test that on a debian aarch64 VM here. The picolibc bits worked correctly, which means I think this is ready to merge whenever the SDK is also ready.

configs/arm-zephyr-eabi.config
```diff
@@ -61,3 +61,10 @@ CT_LIBC_NEWLIB_NANO_ENABLE_TARGET_OPTSPACE=y
 # CT_LIBC_NEWLIB_NANO_LTO is not set
 CT_LIBC_NEWLIB_NANO_NANO_MALLOC=y
 CT_LIBC_NEWLIB_NANO_NANO_FORMATTED_IO=y
+CT_COMP_LIBS_PICOLIBC=y
+CT_PICOLIBC_SRC_CUSTOM=y
+CT_PICOLIBC_CUSTOM_LOCATION="${GITHUB_WORKSPACE}/picolibc"
+CT_LIBC_PICOLIBC_GLOBAL_ATEXIT=y
+CT_LIBC_PICOLIBC_EXTRA_SECTIONS=y
+CT_LIBC_PICOLIBC_EXTRA_CONFIG_ARRAY="-Dsysroot-install=true -Dsysroot-install-skip-checks=true -Dsysroot-install-skip-checks=true"
```
stephanosio, line 69:
```suggestion
CT_LIBC_PICOLIBC_EXTRA_CONFIG_ARRAY="-Dsysroot-install=true -Dsysroot-install-skip-checks=true"
```

stephanosio (commented):
It looks like this PR is not building picolibc for the following targets:

* arc64-zephyr-elf
* nios2-zephyr-elf
* x86_64-zephyr-elf
* xtensa-*_zephyr-elf

I recall that it was a deliberate decision to drop `nios2-zephyr-elf` because there were some issues and the arch is effectively dead so it was not worth the effort.

What about the rest?

keith-packard:
>     * arc64-zephyr-elf

Requires setjmp/longjmp bits which picolibc doesn't have yet.

>     * nios2-zephyr-elf

Skipped as not interesting anymore.

>     * x86_64-zephyr-elf

Needs some fixes in picocrt code (or just skipping it, maybe?)

>     * xtensa-*_zephyr-elf

I'm not sure what the state of these architectures are in picolibc, will need to go investigate.

keith-packard:
> > * arc64-zephyr-elf
> 
> Requires setjmp/longjmp bits which picolibc doesn't have yet.

I found the ARC ABI specs and it looks like the arc32 setjmp/longjmp bits present in picolibc will
work fine -- they save one extra register (R13), which will be harmless.

keith-packard:
> > * x86_64-zephyr-elf
> 
> Needs some fixes in picocrt code (or just skipping it, maybe?)

Nope -- needed to fix stuff so that multilib x86 builds worked at all. Upstream PR is pending, when that works, I'll update picolibc bits and enable x86 here.

configs/arm-zephyr-eabi.config
keith-packard: thanks, a wild sed -i operation went astray.

keith-packard:
>     * xtensa-*_zephyr-elf

For xtensa to work, we need setjmp/longjmp so we can have libstdc++ built. Does anyone have the ABI spec for these chips who can make that happen? Alternatively, we could disable libstdc++ in the SDK.

keith-packard:
> For xtensa to work, we need setjmp/longjmp so we can have libstdc++ built. Does anyone have the ABI spec for these chips who can make that happen? Alternatively, we could disable libstdc++ in the SDK.

Bits merged from the newlib-esp32 project into picolibc. Xtensa libraries are now being built with the rest of the toolchains.

keith-packard:
>     * nios2-zephyr-elf

Picolibc already had nios2 support, it just needed fixing for multilib configurations to work around the gcc flags for FPU configuration which forcibly set -fsingle-precision-constants, breaking the double precision math functions. I've hacked around that and pushed changes to picolibc upstream for merging there. Once merged, I'll add that to his PR.

keith-packard:
Nios2 added; "should" build cleanly.

stephanosio (approved):

PR #294, opened 2020-11-18
gcc: update multilibs for SPARC

martin-aberg:
Successfully tested with Zephyr sanitycheck on the SPARC board configurations `qemu_leon3`, `gr716a_mini` and `generic_leon3`.

A summary of the differences between `-mcpu` options are given below.

| Option                              | mul/div | casa |
| ----------------------------- |  --- | --- |
| -mcpu=v7 (or no -mcpu=) |        NO     |  NO |
| -mcpu=leon                      | YES   |   NO |
| -mcpu=leon3                    | YES   |   YES |
| -mcpu=leon3v7                 |  NO  |     YES |

`mul/div` means that the compiler will generate multiplication and
division instructions, which is a SPARC V8 feature. There are some
use cases for compiling without `mul/div`:
- If the hardware is synthesized without `mul/div` instructions.
- To reduce interrupt response time on systems with slow `div`.

`casa` means that the compiler generates the atomic compare-and-swap
instruction which is an extension to SPARC V8.

`-mcpu=leon` is equivalent to `-mcpu=v8` except the instruction timing.
`-mcpu=leon` is useful for LEON2 systems. `-mcpu=leon3` is the typical
option for LEON3, LEON4 and LEON5 systems.

The `-mflat` option enables the flag register window model. `SAVE` and
`RESTORE` instructions will not be generated by the compiler. `-mflat` is
ABI compliant with the SPARC ABI.

### DIFF
--- a/patches/gcc/10.2.0/0015-Add-LEON-multilibs-for-sparc-zephyr-elf.patch
+++ b/patches/gcc/10.2.0/0015-Add-LEON-multilibs-for-sparc-zephyr-elf.patch
@@ -1,0 +1,96 @@
+From 4beadbc1abc6a2943b541124d391ceb26a1042c4 Mon Sep 17 00:00:00 2001
+From: Martin Aberg <maberg@gaisler.com>
+Date: Wed, 18 Nov 2020 13:18:59 +0100
+Subject: [PATCH] Add LEON multilibs for sparc-zephyr-elf
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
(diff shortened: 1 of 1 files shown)


galak:
LGTM.  Will wait for CI to finish.

martin-aberg:
Thank you @galak for the quick review and processing.

I have can confirm that the binary distribution `zephyr-sdk-0.12.0-beta-2-x86_64-linux-setup.run` works very well.

galak:
> I have can confirm that the binary distribution `zephyr-sdk-0.12.0-beta-2-x86_64-linux-setup.run` works very well.

Thanks!

PR #316, opened 2021-01-04
cmake: add TOOLCHAIN_SUPPORTS_SOFT_FLOAT for x86

dcpleung:
This enables a hidden option to indicate that the toolchain
support software floating point operations.

### DIFF
--- a/cmake/zephyr/Kconfig
+++ b/cmake/zephyr/Kconfig
@@ -6,3 +6,11 @@
 config TOOLCHAIN_ZEPHYR_0_12
 	def_bool y
 	select HAS_NEWLIB_LIBC_NANO if (ARC || (ARM && !ARM64) || RISCV)
+
+config TOOLCHAIN_SUPPORTS_SOFT_FLOAT
+	bool
+	default y if X86 || !X86_64
+	help
+	  Hidden option to indicate whether the toolchain supports software
+	  floating point operations that can be enabled via "-msoft-float"
+	  compiler flag.


dcpleung:
This is needed for https://github.com/zephyrproject-rtos/zephyr/pull/30620

galak:
I wonder if this is better to do on the Zephyr side instead of on the SDK side.  @tejlmand thoughts?

dcpleung:
> I wonder if this is better to do on the Zephyr side instead of on the SDK side. @tejlmand thoughts?

Note that it is hard-coded to source the Kconfig inside SDK in file `cmake/toolchain/zephyr/generic.cmake`.

tejlmand:
> I wonder if this is better to do on the Zephyr side instead of on the SDK side. @tejlmand thoughts?

It seems to me that this is dependent on how the toolchain itself was built.
A standard gcc toolchain for x86 will support this, but other gcc variant may not, as example the  gnuarmemb does not support this flag.

But I assume Zephyr SDK gcc for x86 will always support this, but then we would start to have Zephyr SDK Kconfig settings / knowledge both in https://github.com/zephyrproject-rtos/sdk-ng/blob/master/cmake/zephyr/Kconfig and somewhere in https://github.com/zephyrproject-rtos/zephyr/tree/master/cmake/toolchain/zephyr.
And do we want that ?
What we don't want, is to go back to a `${ZEPHYR_BASE}/cmake/toolchain/zephyr/X.Y` situation, cause we know what that will lead to.

galak:
> But I assume Zephyr SDK gcc for x86 will always support this, but then we would start to have Zephyr SDK Kconfig settings / knowledge both in https://github.com/zephyrproject-rtos/sdk-ng/blob/master/cmake/zephyr/Kconfig and somewhere in https://github.com/zephyrproject-rtos/zephyr/tree/master/cmake/toolchain/zephyr.
> And do we want that ?
> What we don't want, is to go back to a `${ZEPHYR_BASE}/cmake/toolchain/zephyr/X.Y` situation, cause we know what that will lead to.

Didn't we discuss in the toolchain working group adding Kconfig or Cmake info for various toolchain features -- for example if newlib was supported or not?? 

I was thinking having whatever Kconfig symbols we add be on the Zephyr side so as to not require additional releases of the SDK whenever we need a new one for something.  We can have those Kconfig symbols be:

```
if TOOLCHAIN_ZEPHYR_0_12

config TOOLCHAIN_SUPPORTS_SOFT_FLOAT
   bool
   default y if X86 || !X86_64
...
endif
```

tejlmand:
> Didn't we discuss in the toolchain working group adding Kconfig or Cmake info for various toolchain features -- for example if newlib was supported or not??

we did, and this might be one more to consider.

For this setting we need to be sure it is a toolchain thing or if it is purely the compiler / target thing.
From discussions in https://github.com/zephyrproject-rtos/zephyr/pull/30620 it starts to look as this is just gcc, and not a Zephyr SDK specific thing. (when saying gcc, it means the flag exists for other gcc based compilers, like gnuarmemb)

dcpleung:
No longer needed.

PR #325, opened 2021-03-26
cmake: Add arm64 cross-compile target

carlocaione:
There is an ongoing effort in Zephyr to separate the ARM architecturefrom ARM64 (AArch64). As part of this work, a new architecture will be introduced called 'arm64'.

To be able to keep using the same SDK to compile old and new Zephyr releases, we need to add a new cross-compile target specifically for 'arm64' together with the legacy 'arm' target.

@galak @tejlmand  this should be a pretty harmless change to be able to support the current situation and at the same time the new changes that I'm going to submit soon, where the ARM64 will be separated from the ARM architecture.

### DIFF
--- a/cmake/zephyr/target.cmake
+++ b/cmake/zephyr/target.cmake
@@ -1,10 +1,12 @@
 # SPDX-License-Identifier: Apache-2.0
 
 if(CONFIG_ARM64)
+  # Legacy target to support Zephyr versions before the ARM/ARM64 split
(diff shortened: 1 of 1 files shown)


galak (approved):

carlocaione:
I think the buildkite failure is expected.

galak:
> I think the buildkite failure is expected.

It is.  Would like @tejlmand review, and than I'll merge.

tejlmand (approved):

PR #353, opened 2021-07-07
config: xtensa: add esp32 config and overlay

sylvioalves:
Current implementation uses GCC and binutils
provided by external espressif repository as
patches have not been subbmitted upstream.

Signed-off-by: Sylvio Alves <sylvio.alves@espressif.com>


configs/xtensa_esp32.config
```diff
@@ -0,0 +1,102 @@
...
+# CT_LOG_PROGRESS_BAR is not set
+CT_TARGET_CFLAGS="-mlongcalls"
+CT_ARCH_XTENSA=y
+CT_XTENSA_CUSTOM=y
+CT_OVERLAY_NAME="esp32"
+CT_OVERLAY_LOCATION="./overlays"
+CT_TARGET_VENDOR="esp32"
+CT_NEWLIB_SRC_DEVEL=y
+# CT_NEWLIB_DEVEL_URL="https://github.com/zephyrproject-rtos/newlib-cygwin.git"
+# CT_NEWLIB_DEVEL_REVISION="4f5997d3c0f9011135e9627dad700c9d64be4a4b"
```
sylvioalves, line 16: I will fix this when merged.

galak:
So I have a general concern here if the various toolchain components are in sync with the other Zephyr toolchains.  Ie, gcc, binutils, gdb versions.  The general rule of thumb I've had is that all the arch's should be using the same version/source basis to build from.

If that isn't feasible for esp32 support, I'm not sure what value we get in supporting this in the Zephyr SDK vs just pointing at the esp32 toolchains you guys already have.

stephanosio:
> So I have a general concern here if the various toolchain components are in sync with the other Zephyr toolchains. Ie, gcc, binutils, gdb versions. The general rule of thumb I've had is that all the arch's should be using the same version/source basis to build from.
> 
> If that isn't feasible for esp32 support, I'm not sure what value we get in supporting this in the Zephyr SDK vs just pointing at the esp32 toolchains you guys already have.

I agree with @galak here.

I assume this PR is still a draft. As discussed offline, all ESP32-specific patches should either be added here (in case binutils-gdb and gcc) or merged into the relevant Zephyr fork (in case of newlib).

sylvioalves:
@galak @stephanosio sure, this a WIP in fact. I agree with you and I have been checking gcc and gdb patches to overcome this.

Issue #349, opened 2021-06-07
crosstool-ng does not strip nano variant libs


stephanosio:
Zephyr crosstool-ng does not strip the nano variant libs (`libc_nano.a`, `libstdc++_nano.a`, ...).

The commit that added this feature is: https://github.com/zephyrproject-rtos/crosstool-ng/commit/5dc17d3bb1781269b7594cd643a32b316ffcef3c

It is likely that this feature initially worked, but later stopped working after updating to the latest crosstool-ng with different nano variant lib support patches merged upstream.

stephanosio:
cc @galak

stephanosio:
It looks like I've already fixed this earlier:
https://github.com/zephyrproject-rtos/crosstool-ng-old/commit/8c79efc88232af6d51b3a9d9c95058e0fd606ff4

galak:
I pulled that commit into crosstool-ng SDK-0.13.0-test

galak:
PR #356

galak:
@stephanosio can you send a upstream PR for this to crosstool-ng

stephanosio:
@galak will do, once we verify that it works in that PR.

stephanosio:
With https://github.com/zephyrproject-rtos/crosstool-ng-old/commit/8c79efc88232af6d51b3a9d9c95058e0fd606ff4 (built by #356), the library size looks more sane now:
```
-rw-r--r-- 2 stephanos stephanos 1.3M Jul 15 01:21 aarch64-zephyr-elf/aarch64-zephyr-elf/lib/libc.a
-rw-r--r-- 1 stephanos stephanos 1.4M Jul 15 01:21 aarch64-zephyr-elf/aarch64-zephyr-elf/lib/libc_nano.a
-rw-r--r-- 1 stephanos stephanos 4.6M Jul 15 01:21 aarch64-zephyr-elf/aarch64-zephyr-elf/lib/libstdc++_nano.a
-rw-r--r-- 1 stephanos stephanos 4.9M Jul 15 01:21 aarch64-zephyr-elf/aarch64-zephyr-elf/lib/libstdc++.a
-rw-r--r-- 2 stephanos stephanos 1.3M Jul 15 02:12 arc64-zephyr-elf/arc64-zephyr-elf/lib/libc.a
[...]

PR #356, opened 2021-07-14
crosstool-ng: pull in fix to strip nano libs

galak:


stephanosio:
@galak For 0.13.0, this should do; but, for upstreaming to crosstool-ng, these patches need to be reworked.
See https://github.com/zephyrproject-rtos/sdk-ng/issues/349#issuecomment-882027711

stephanosio (approved):

PR #357, opened 2021-07-14
fixup scripts to allow pkg on macosx

galak:


scripts/template_dir
```diff
@@ -33,13 +33,15 @@ verify_os () {
 case "$(uname -s)" in
 
    Darwin)
-     echo 'OS X is not supported!'
-     exit 1
+     MD5=md5
+     os=macosx
      ;;
 
    # Linux, the only supported OS
```
stephanosio, line 40: Should remove this now that macOS is supported (sort of).

scripts/make_zephyr_sdk.sh
```diff
@@ -21,11 +21,25 @@ fi
 root_dir=$(dirname $0)/..
 sdk_version=$(cat $root_dir/VERSION)
 machine=$1
+
+case "$(uname -s)" in
+   Darwin)
+     os=macosx
```
stephanosio, line 27: would probably be better to name it `macos` since "Mac OS X" nomenclature is no longer used (and more importantly, it isn't "X," aka. 10, anymore).

stephanosio (approved):

PR #359, opened 2021-07-18
ci: Add workflow to build macosx-x86_64 toolchains

galak:
Add a github workflow that will build macosx-x86_64 toolchains and
publish the packages on release.  The workflow only runs after a
commit since it takes 6+ hours to run on current github macosx
builders.

A few notes / caveats on the macosx toolchain build / packages:

* NO host tools (ie qemu, openocd, dtc, etc).  Since we dont build
  yocto on MacOS we dont currently have any of the host tools at
  this time

* GDB build is without python.  The python GDB on MacOS doesn't
  currently work so we limit GDB to only a non-python enabled
  build.

* gcc build with --without-zstd to make builds portable.  As zstd
  isn't a common library on MacOS we build with this flag since
  it is found in homebrew with the github MacOS setup

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -1,0 +1,227 @@
+name: macos-ci
+
+on:
+  push:
+
+jobs:
+  # Cancel job is setup to not cancel jobs associated with
+  # a tag (ie a release).  Also we run it on ubuntu in the
+  # hopes that if the macos builders are full that it can
+  # still run
+  cancel:
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: "Cancel Previous Runs"
+        uses: styfle/cancel-workflow-action@0.9.0
+        if: ${{ !startsWith(github.ref, 'refs/tags/') }}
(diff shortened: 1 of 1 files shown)


stephanosio (commented):
> GDB build is without python. The python GDB on MacOS doesn't
currently work so we limit GDB to only a non-python enabled
build.

If I remember correctly, this used to work. Is it a build failure or a run-time one?

galak:
> If I remember correctly, this used to work. Is it a build failure or a run-time one?

Its a runtime failure.  Not sure what exactly is going on.

galak:
@stephanosio I would like to merge this and we can look at fixing up the issues w/GDB and such as future items.  Let me know if you've got any other concerns.

stephanosio:
> @stephanosio I would like to merge this and we can look at fixing up the issues w/GDB and such as future items. Let me know if you've got any other concerns.

@galak sounds fair.

stephanosio (approved):

PR #362, opened 2021-07-21
scripts/template_dir: Add file exist checks

galak:
The current MacOS builds do not have environment-setup* files
or version-* that come from the Yocto hosttools builds.  So add
some checks in template_dir to see if those files exist before
trying to remove or move them.

### DIFF
--- a/scripts/template_dir
+++ b/scripts/template_dir
@@ -123,10 +123,16 @@ do_cleanup()
 	cd $target_sdk_dir
 	if [ $post_install_cleanup = "1" ]; then
 		#rm site-config*
-		rm environment-setup*
+		env_files=(environment-setup*)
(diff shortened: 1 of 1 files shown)


scripts/template_dir
```diff
@@ -123,10 +123,16 @@ do_cleanup()
 	cd $target_sdk_dir
 	if [ $post_install_cleanup = "1" ]; then
 		#rm site-config*
-		rm environment-setup*
+		env_files=(environment-setup*)
+		if [ -e "${env_files[0]}" ]; then
+			rm environment-setup*
+		fi
```
stephanosio, line 129: Could probably just do `rm -f` but it doesn't really matter.

stephanosio (approved):

PR #377, opened 2021-07-28
newlib: workaround cflags issue in crosstool for newlib-nano

galak:
When building newlib-nano because its treated as a companion_libs the
order it gets built ends up not having CT_ARCH_TARGET_CFLAGS set
correctly (as it gets cleared by a previous build stage).

This causes issues on ARM, so to workaround this we mimic the setting
of CT_ARCH_TARGET_CFLAGS in CT_LIBC_NEWLIB_NANO_TARGET_CFLAGS to
"-mthumb-interwork" (That mimics how CT_ARCH_TARGET_CFLAGS is set
because we enable CT_ARCH_ARM_INTERWORKING=y).


galak:
Not sure if this will fix the issue, but doing a build to see what we get.

PR #378, opened 2021-07-28
crosstool-ng: Update to pull in fix for nano.spec

### LINKED ISSUE #369: Footprint increase
A significant footprint increase has been reported for the SDK 0.13.0-* releases.

For `samples/net/sockets/coap_client` with `nrf52840dk_nrf52840`

**0.12.4**
```
           FLASH:      116016 B         1 MB     11.06%
            SRAM:       28320 B       256 KB     10.80%
        IDT_LIST:          0 GB         2 KB      0.00%
```

**0.13.0-rc3**
```
           FLASH:      143080 B         1 MB     13.65%
            SRAM:       30336 B       256 KB     11.57%
        IDT_LIST:          0 GB         2 KB      0.00%
```

This increase happened after the commits in the PR #333, which switched to using the upstream crosstool-ng newlib nano build process.

galak:
pull in nano.spec fix that links with -lc_nano instead of -lc (same
for other nano lib variants, libm, libstdc++, etc..)

Fixes #369

### DIFF
--- a/go.sh
+++ b/go.sh
@@ -28,7 +28,7 @@ if [ "$os" == "macos" ]; then
 	SDK_NG_HOME="/Volumes/${ImageName}"
 fi
 
-CROSSTOOL_COMMIT="4d5660e34e7c5f522e8b07ded82d7a6a15b787ef"
+CROSSTOOL_COMMIT="e0117582601a9a60fe279fda920da3875a0db366"
 build_crosstool()
 {
 	# only build if we don't already have a built binary


go.sh
```diff
@@ -176,7 +176,7 @@ for t in ${TARGETS}; do
 	rm -rf ${CT_PREFIX}/*/newlib-nano
```
stephanosio, line 176: We should just set `CT_NEWLIB_NANO_INSTALL_IN_TARGET=n` and use the separate `newlib-nano` directory as done in Keith's original implementation. That seems to be a much cleaner approach IMO.

stephanosio (changes requested):
This does fix the problem at hand, but I think we could probably do something else/arguably better.

See https://github.com/crosstool-ng/crosstool-ng/pull/1568#pullrequestreview-717613082

go.sh
stephanosio: This was evaluated in #379, but this does not seem feasible for the time being (see https://github.com/zephyrproject-rtos/sdk-ng/pull/379#issuecomment-889252433).

Reconsider switching to #379 in the future when the target library stripping patch is updated to handle stripping the libraries under the `newlib-nano` directory.

stephanosio:
Confirmed working:
```
           FLASH:      116016 B         1 MB     11.06%
            SRAM:       28320 B       256 KB     10.80%
        IDT_LIST:          0 GB         2 KB      0.00%
```

stephanosio (approved):

PR #397, opened 2021-09-17
0.13.1

galak:


stephanosio:
@galak Are we going to use this for 2.7? or is this for the future Zephyr releases?

galak:
> @galak Are we going to use this for 2.7? or is this for the future Zephyr releases?

For 2.7, I don't see any reason not to make this available.  People can still use SDK 0.13.0 if they want.

stephanosio (approved):

stephanosio:
> For 2.7, I don't see any reason not to make this available. People can still use SDK 0.13.0 if they want.

shouldn't be a problem as long as we do not update the CI to use this right now.

galak:
> shouldn't be a problem as long as we do not update the CI to use this right now.

Lets discuss in the toolchain meeting, as I think we should update v2.7 to use this because of the ARM CVE fix that is pulled in.

Issue #399, opened 2021-10-02
Mirror/cache sources required to build SDK


stephanosio:
As part of the Zephyr SDK build process, crosstool-ng downloads the source tarballs for the required components from external servers (mirror URLs specified under the crosstool-ng `packages/`).

This makes our CI infrastructure depend on the external servers and often leads to CI failures when these servers go down for whatever reason.

Note that we download the source tarballs from many different servers hosted by many different projects/organisations, so the chances of any one of them being down are not so unlikely; in fact, we have seen the failures related to this many times so far.

For instance, in https://buildkite.com/zephyr/sdk-ng/builds/462, CI failed due to the ISL component (required to build gcc) mirror being unavailable:
```
[INFO ]  Retrieving needed toolchain components' tarballs
...
[EXTRA]    Retrieving 'isl-0.24'
[ERROR]    isl: download failed
```

In order to ensure that these failures do not occur, we must mirror the source tarballs for the required components in a server/service managed by or affiliated with the Zephyr Project.

stephanosio:
@galak @nashif I think it would be the most hassle-free to just set up an AWS S3 bucket for this. What do you think?

galak:
> @galak @nashif I think it would be the most hassle-free to just set up an AWS S3 bucket for this. What do you think?

No issues with the general idea, just need to see how this would work.

stephanosio:
TODO: Apply the same strategy for the new CI (#417).

stephanosio:
Consider caching source files in the [docker-sdk-build](https://github.com/zephyrproject-rtos/docker-sdk-build) CI image.

PR #402, opened 2021-10-04
download testing

galak:


galak:
This should address #399

stephanosio:
So these clone failures with `repo.or.cz` happen because they are using the SSL certificates issued by "Let's Encrypt" whose old root certificate expired on 30 Sep. 2021, and the CI image, which is based on Ubuntu 18.04 snapshot on 17 Jul. 2020, does not contain the new root certificate.

@galak The fix can be as simple as releasing a new [`docker-sdk-build`](https://github.com/zephyrproject-rtos/docker-sdk-build) image with an updated `ca-certificates` package.

galak:
> @galak The fix can be as simple as releasing a new [`docker-sdk-build`](https://github.com/zephyrproject-rtos/docker-sdk-build) image with an updated `ca-certificates` package.
> 
> p.s. of course, we can just do `apt update` and `apt install ca-certificates` in the CI scripts as a temporary hack too.

So do you think just "rebuilding" the docker image will be sufficient?

stephanosio:
> So do you think just "rebuilding" the docker image will be sufficient?

well, that really depends depends on how old the base `ubuntu:18.04` image is.

https://github.com/zephyrproject-rtos/docker-sdk-build/blob/2d1fb740748329012b1209cd3b43e048d497f7f0/setup.sh#L4
It would be good to add `apt-get upgrade -y` or `apt-get install -y ca-certificates` (which will install a newer version of the `ca-certificates` package if there is one) just in case.

PR #404, opened 2021-10-08
meta-zephyr-sdk: xilinx_qemu: Update to QEMU 5.1.0

pfalcon:
Uses the same git revision as Xilinx PetaLinux 2021.1. Allows to be better
aligned with vendor tooling/samples/etc. when working on integration
projects like OpenAMP.

Signed-off-by: Paul Sokolovsky <paul.sokolovsky@linaro.org>


pfalcon:
> All kernel tests are passing for qemu_cortex_r5 with this change:

@stephanosio: Thanks for confirming that!

PR #411, opened 2021-11-14
config: Disable the GP register optimizations on MIPS

remy-luisant:
This change turns off the use of the GP register for small data values on MIPS.
This allows a few more testcases to pass.

Signed-off-by: Remy Luisant <remy@luisant.ca>

### DIFF
--- a/configs/mips.config
+++ b/configs/mips.config
@@ -5,6 +5,7 @@ CT_LOCAL_TARBALLS_DIR="${CT_PREFIX:-${HOME}/x-tools}/sources"
 # CT_LOG_PROGRESS_BAR is not set
 CT_ARCH_MIPS=y
 CT_MULTILIB=y
+CT_TARGET_CFLAGS="-G0 -mno-gpopt"
 CT_TARGET_VENDOR="zephyr"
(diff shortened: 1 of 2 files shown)


stephanosio (commented):
> This change turns off the use of the GP register for small data values on MIPS.
This allows a few more testcases to pass.

Could you provide a bit more detailed explanation in the commit message as to why this is supposed to work?

remy-luisant:
@stephanosio 

I have done as you requested.

stephanosio:
@remy-luisant Thanks.

> Compilers on MIPS customarily use a dedicated global register for access to a
pool of small data items. This is not currently supported by the MIPS port.

Just curious, are there any technical reasons why this feature is not currently supported by the MIPS port? and do you plan to support this feature eventually?

remy-luisant:
@stephanosio 

There is no technical reason. A version of the MIPS port that has the optimization enabled exists but fails many testcases. I spent some time investigating the failures, but they don't occur in temporal proximity to the uses of the gp register, which makes diagnosing not trivial. Given that the gains from having the optimization enabled are minimal, the priorities lie elsewhere.

stephanosio (approved):

PR #422, opened 2021-12-26
qemu: Add second UART to ARC's "virt" platform

abrodkin:
This enables more use-cases, including separation of debug
output from interaction with the platform via serial port,
SLIP-based networking and probably much more.

Originally introduced here: https://github.com/foss-for-synopsys-dwc-arc-processors/qemu/pull/62
Requested here: https://lists.zephyrproject.org/g/devel/message/8285


jobythampan (approved):
I tried this and it works

stephanosio:
This will be addressed as part of #445.

PR #444, opened 2022-03-15
Add toolchain components as Git submodules

stephanosio:
This series adds the toolchain components (e.g. binutils, gcc, ...) as Git submodules and updates the crosstool-ng config files to use them.

This allows the toolchain components to be managed in a more centralised manner.


stephanosio:
Re: SRC_CUSTOM

https://github.com/crosstool-ng/crosstool-ng/commit/f17fba608042e504f8506d51699dae8acc9abcea

abrodkin:
@stephanosio I'd suggest to remove all the branches in Zephyr components forks except related to Zephyr, otherwise it's really confusing:
1. I guess we're not going to update most of the branches from upstream anyway, even though in upstream they might change. Think of "stable" branches like `releases/gcc-10`, `releases/gcc-11` in GCC.
2. It's hard to find those really related to Zephyr.

stephanosio:
> @stephanosio I'd suggest to remove all the branches in Zephyr components forks except related to Zephyr, otherwise it's really confusing:

@abrodkin Indeed. I am planning to tidy these up soon.

PR #458, opened 2022-04-05
Add minimal distribution bundle

### LINKED ISSUE #450: Support web-based install from minimal distribution bundle
Support web-based installation of the Zephyr SDK from a minimal distribution bundle that only contains the bare minimum to bootstrap the setup process.

The following changes need to be implemented:

1. Update the distribution bundle setup script to support downloading and installing individual toolchain and host tool components from the GitHub release assets based on user's request.
2. Create a separate minimal distribution bundle containing the "web-based" setup script.

Note that the "minimal distribution bundle" will be distributed alongside the current "full distribution bundle" which contains everything. User will be able to choose between these two on the download page.

stephanosio:
This series adds the changes to support building the "minimal distribution bundle."

Closes #450

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1281,9 +1281,9 @@ jobs:
 
         if [ "${{ runner.os }}" == "Windows" ]; then
           # NOTE: Escape forward slashes because MinGW (bash)
-          ./setup.cmd //c //t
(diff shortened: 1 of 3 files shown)


stephanosio:
Tested in https://github.com/zephyrproject-rtos/sdk-ng-testing/actions/runs/2106308785

Issue #468, opened 2022-05-05
Incorrect newlib.h is used when nano.specs is used


stephanosio:
`newlib.h` is responsible for defining the newlib build configurations so that the newlib headers can expose only the enabled features to the application code, and the `nano.specs` is responsible for configuring the include and library paths for the nano build of the newlib.

When `--specs=nano.specs` is specified to GCC, the newlib nano (`libc_nano.a`) is linked instead of the full newlib (`libc.a`) and the `include/newlib-nano` directory, which contains the `newlib.h` specific to the newlib nano build, is added as a system include directory, such that the nano `newlib.h` is included by application code instead of the full `newlib.h`, which resides in `include`.

The `nano.specs` file currently specifies the `newlib.h` include path as follows:

```
*cpp_unique_options:
-isystem =/include/newlib-nano %(nano_cpp_unique_options)
```

This is incorrect -- the `=` at the beginning of the include path is effectively no-op and GCC tries to literally resolve the path as `=/include/newlib-nano`, which obviously is not a valid path, and this leads to this include path being ignored and the `newlib.h` for the full newlib build from `include/` being included by the application code:

```
$ echo | /opt/zephyr-sdk-0.14.1/arm-zephyr-eabi/bin/arm-zephyr-eabi-gcc --specs=nano.specs -E -Wp,-v -
ignoring nonexistent directory "=/include/newlib-nano"
ignoring duplicate directory "/opt/zephyr-sdk-0.14.1/arm-zephyr-eabi/bin/../lib/gcc/../../lib/gcc/arm-zephyr-eabi/10.3.0/include"
ignoring duplicate directory "/opt/zephyr-sdk-0.14.1/arm-zephyr-eabi/bin/../lib/gcc/../../lib/gcc/arm-zephyr-eabi/10.3.0/include-fixed"
ignoring duplicate directory "/opt/zephyr-sdk-0.14.1/arm-zephyr-eabi/bin/../lib/gcc/../../lib/gcc/arm-zephyr-eabi/10.3.0/../../../../arm-zephyr-eabi/sys-include"
ignoring duplicate directory "/opt/zephyr-sdk-0.14.1/arm-zephyr-eabi/bin/../lib/gcc/../../lib/gcc/arm-zephyr-eabi/10.3.0/../../../../arm-zephyr-eabi/include"
#include "..." search starts here:
#include <...> search starts here:
[...]

PR #469, opened 2022-05-09
Fix newlib include path and always enable C99 format specifier support

stephanosio:
```
This commit enables the C99 format specifier support for the newlib
nano variant since it is required for C99 standard compatibility.

Without this, Zephyr applications cannot make use of the format
specifiers newly added in the C99 standard such as `%hhu` and `%hhd`.

Moreover, the newlib `inttypes.h` defines `PRI*8` macros as `%hh*` even
when this configuration is disabled, effectively making the abstraction
provided by the `inttypes.h` useless due to lack of universality; for
this reason, the C99 format specifier support must always be enabled
for both full and nano variants of the newlib.

For more details, refer to the issue https://github.com/zephyrproject-rtos/zephyr/issues/45336.

In addition, this commit disables the nano-formatted I/O build option
for the newlib nano variant because this option limits the formatted
I/O features to that of the C89 standard and overrides the C99 format
specifier support option -- refer to the newlib documentation of the
`--enable-newlib-nano-formatted-io` configuration.

```

```
    crosstool-ng: Pull in newlib nano.specs include path patch

    This commit pulls in the crosstool-ng patch to fix the incorrect
    newlib-nano include path specified by the toolchain `nano.specs` file.

    For more details, refer to the issue #468.

```

crosstool-ng PR: https://github.com/zephyrproject-rtos/crosstool-ng/pull/16

Fixes https://github.com/zephyrproject-rtos/sdk-ng/issues/468
Fixes https://github.com/zephyrproject-rtos/zephyr/issues/45336


stephanosio:
DNM until https://github.com/zephyrproject-rtos/crosstool-ng/pull/16 is merged and the submodule commit is updated to that of the merged commit.

stephanosio:
TODO: Check and document the footprint increase in the newlib-nano variant after enabling the C99 format specifier support.

Even if the footprint increase is significant however, as noted in https://github.com/zephyrproject-rtos/zephyr/issues/45336, we must proceed with this change because this is required for us to be C99 compliant and universally make use of the C99 inttypes in the Zephyr codebase:

> we should not be optimising the newlib footprint at the cost of breaking C99 compatibility.

stephanosio:
Building a modified `samples/hello_world` with `printf` instead of `printk` for `qemu_cortex_m3`:

_Newlib-nano, before this PR_
```
           FLASH:       14064 B       256 KB      5.36%
            SRAM:        4120 B        64 KB      6.29%
        IDT_LIST:          0 GB         2 KB      0.00%
```

_Newlib-nano, after this PR_
```
           FLASH:       29236 B       256 KB     11.15%
            SRAM:        4480 B        64 KB      6.84%
        IDT_LIST:          0 GB         2 KB      0.00%
```

Note that, by enabling the C99 format specifier support in the newlib nano variant, the flash footprint almost doubled (increased from 14064 bytes to 29236 bytes).

While this may seem like a very bad outcome at first glance, as noted earlier, this is a necessary change for us to be able to universally make use of the C99 format specifiers in the Zephyr codebase, which is necessary for C99 compliance as well as writing and supporting portable code.

To reiterate the statement made in the original issue:

> we should not be optimising the newlib footprint at the cost of breaking C99 compatibility.

We could potentially further reduce the newlib nano footprint, even with the C99 format specifier support enabled, by making a number of breaking code changes to the newlib, but these changes are likely never going to be accepted by the upstream newlib project, and will be very high maintenance in the long run.

As noted in https://github.com/zephyrproject-rtos/zephyr/issues/45336, with the introduction of the picolibc as a replacement for the minimal libc, the lowest end requirement of the newlib nano C library can be loosened, so the aforementioned effort to further reduce the newlib nano footprint with the C99 format specifier support enabled may not be worth the investment.

For these reasons, in spite of the significant footprint increase, we are going to proceed with this change so that the Zephyr codebase can readily make use of the C99 features and be free of the kind of issues observed in #16645.

As an additional note, the newlib nano variant footprint is still considerably smaller than that of the newlib full variant even with the C99 format specifier support enabled, so the effectiveness of the nano variant is by no means nullified by this change.

_Newlib-full_
```
           FLASH:       38624 B       256 KB     14.73%
            SRAM:        6112 B        64 KB      9.33%
        IDT_LIST:          0 GB         2 KB      0.00%
```

PR #476, opened 2022-05-13
meta-zephyr-sdk: arc_qemu: Update to 2022.04.26 release

stephanosio:
This commit updates the ARC QEMU to 2022.04.26 release (tag).

NOTE: This change was initiated because the original commit
2625a6fbc3f60233e1d340931d00242d82230091 no longer exists in the
upstream repository.


abrodkin (approved):
Wow, I was meant to create that same PR for quite some time (see I even created a tag with verified solution for https://github.com/foss-for-synopsys-dwc-arc-processors/qemu/issues/67).

Thanks for taking care of that!

PR #506, opened 2022-06-24
overlays: Convert Xtensa overlays to directory

stephanosio:
crosstool-ng allows specifying a toolchain overlay as a directory.

This commit extracts contents of the toolchain overlay archive file
into the corresponding directory so that any changes to the overlay
itself can be tracked in the version control system.


stephanosio:
re: crosstool-ng supporting directories as overlays: https://github.com/crosstool-ng/crosstool-ng/blob/4ad0620fcdd8e7ff3148286e7ea2d431e9546176/scripts/functions#L2255-L2256

PR #509, opened 2022-06-29
gcc: Pull in RISC-V bit manipulation extension multilib patch

stephanosio:
This commit pulls in the GCC patch that builds the multilibs for the
RISC-V target configurations that support the bit manipulation
extension (Bitmanip).


stephanosio:
DNM until https://github.com/zephyrproject-rtos/gcc/pull/6 is merged.

PR #511, opened 2022-07-04
crosstool-ng: Pull in LZMA and Guile scripting support configs

### LINKED ISSUE #505: macOS GDB depends on liblzma.5.dylib
macOS GDB build depends on the `liblzma.5.dylib` dynamic library, which is not available on macOS by default (installed through Homebrew?).

Ensure that GDB builds without any non-default shared/dynamic library dependencies on all supported host operating systems. Note that `libpython3.8` is an exception because it does not make sense to include a Python runtime as part of the Zephyr SDK, and static linking of it does not work.

To be addressed after the upgrade to GDB 12.1 (https://github.com/zephyrproject-rtos/sdk-ng/issues/467).

stephanosio:
This commit pulls in the crosstool-ng patch that adds the
configurations to explicitly control the liblzma and libguile library
linkage so that the shared version of these libraries are not
automatically linked on the systems that have these libraries
installed (e.g. macos-x86_64).

Note that these configurations are disabled by default, so simply
pulling this patch in should be enough to disable the liblzma and
libguile library linkage.

Fixes #505


stephanosio:
DNM until https://github.com/zephyrproject-rtos/crosstool-ng/pull/20 is merged.

PR #531, opened 2022-08-03
ARC: ARCv3: fix toolchain prefix for ARCv3 32bit

evgeniy-paltsev:
Both 32bit and 64bit ARCv3 have same toolchain prefix - `arc64-zephyr-elf`


abrodkin (approved):
Indeed, even though it's not immediately obvious, but both `arc32-*` & `arc64-*` stands for all ARCv3 targets be those 32-bit or 64-bit ones, see https://github.com/foss-for-synopsys-dwc-arc-processors/gcc/blob/arc64/gcc/config.gcc#L354.

stephanosio:
@evgeniy-paltsev Please rebase to fix the CI failure.

PR #538, opened 2022-08-17
ci: Install ninja and meson

stephanosio:
Install ninja and meson, which are required for building the picolibc.


keith-packard (approved):
Yes please!

PR #541, opened 2022-08-19
riscv: Make GCC default architecture include zicsr and zifencei

keith-packard:
This makes the new version of GCC compatible with older versions
when no -march is specified on the command line


stephanosio (approved):
Good catch, thanks.

PR #550, opened 2022-08-29
gcc: arm: Skip FP clobber warning on Cortex-M

microbuilder:
Skips the FP stack clobbering warning when compiling for Cortex-M
targets, which don't have the same stack-handling requirements inside
ISRs.

Signed-off-by: Kevin Townsend <kevin.townsend@linaro.org>


stephanosio:
Rebased

PR #552, opened 2022-08-30
riscv: add library support for RV32IAFC

tari:
IT8xxx2 has an FPU but cannot use -march=rv32imafc: add multilib support
for -march=rv32iafc so the FPU can be used on IT8xxx2. This adds -iafc
as a multilib alias for -iac so doesn't build any additional libraries;
it only teaches the compiler that the -iac libraries can be used for
-iafc systems.

Use of --with-multilib-generator replaces all the default multilib
targets, so the complete set of multilib options are copied from GCC's
defaults in `gcc/config/riscv/t-elf-multilib`.

Signed-off-by: Peter Marheine <pmarheine@chromium.org>

### DIFF
--- a/configs/riscv64-zephyr-elf.config
+++ b/configs/riscv64-zephyr-elf.config
@@ -29,7 +29,7 @@ CT_LIBC_NEWLIB_RETARGETABLE_LOCKING=y
 CT_LIBC_NEWLIB_EXTRA_SECTIONS=y
 CT_GCC_SRC_CUSTOM=y
 CT_GCC_CUSTOM_LOCATION="${GITHUB_WORKSPACE}/gcc"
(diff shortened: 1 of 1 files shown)


stephanosio (changes requested):
This will not work for Zephyr SDK.

Please update https://github.com/zephyrproject-rtos/gcc/blob/zephyr-gcc-12.1.0/gcc/config/riscv/t-zephyr

tari:
Ah, thanks for the pointer.

keith-packard:
I would think you'd want the math libraries built for the FPU, in which case you'd need to add a new multilib configuration. This would also give you FPU exception support and rounding modes.

stephanosio:
New PR: https://github.com/zephyrproject-rtos/gcc/pull/14

PR #554, opened 2022-08-31
cmake: warn users if HOST_TOOLS_HOME is missing

tejlmand:
Warn users if HOST_TOOLS_HOME is missing in a Zephyr SDK installation
as this indicates an incomplete installation.

Signed-off-by: Torsten Rasmussen <Torsten.Rasmussen@nordicsemi.no>

PR #555, opened 2022-08-31
cmake: support loading of host-tool through COMPONENTS

tejlmand:
Support loading of host tools as part of Zephyr-sdk.

This allows users of Zephyr SDK to do:
`find_package(Zephyr-sdk COMPONENTS HostTools)`
instead of having to do something like:
`find_package(Zephyr-sdk COMPONENTS HostTools)`
`include(${ZEPHYR_SDK}/cmake/zephyr/host-tools.cmake)`

it further improve loading of host-tools as those are only loaded if
installed.
Because host tools are now loaded as a find module, then generic CMake
reporting regarding module loading is used.

This allows us to cleanup Zephyr CMake code as well as place the
responsibility of checking and loading of host tools into Zephyr SDK.

Signed-off-by: Torsten Rasmussen <Torsten.Rasmussen@nordicsemi.no>

### DIFF
--- a/cmake/FindZephyr-sdk_HostTools.cmake
+++ b/cmake/FindZephyr-sdk_HostTools.cmake
@@ -1,0 +1,37 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (c) 2022, Nordic Semiconductor ASA
+
+# FindZephyr-sdk_HostTools module for locating HostTools components in Zephyr SDK.
+#
+# The module defines the following variables:
+#
+# 'HOST_TOOLS_HOME'
+# Path to host tools home if installed.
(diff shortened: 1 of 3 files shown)


tejlmand:
alternative to #554

stephanosio:
To be discussed at the Toolchain WG meeting today

stephanosio:
**Toolchain WG (2022-09-06)**

* Since we are making (potentially breaking) changes to the interface between Zephyr SDK and Zephyr, we would want to do it right the first time.
* `COMPONENTS` parameter, although interesting, would not work if we wanted to to release the Zephyr SDK components (toolchains for separate architectures, individual host tools components; as proposed in https://github.com/zephyrproject-rtos/zephyr/issues/37255) because the version requirement can only be specified at the package level.
* In general, this needs more thinking. To be discussed further.

PR #557, opened 2022-09-07
cmake: Declare toolchain support for thread local storage

keith-packard:
The Zephyr SDK toolchains all enable TLS, so add a Kconfig setting which reports this to the Zephyr build. This will allow the explicit zephyr toolchain variant check in Zephyr's kernel/Kconfig to eventually be removed.

This is part of a fix associated with https://github.com/zephyrproject-rtos/zephyr/pull/49087


stephanosio (changes requested):
@keith-packard It looks like this PR contains a spurious update of the `crosstool-ng` submodule.

keith-packard:
oops! Fixed.

stephanosio (approved):

PR #559, opened 2022-09-22
Pull in various GCC updates

stephanosio:
Pull in various GCC updates for ARC and RISC-V.


stephanosio:
DNM until https://github.com/zephyrproject-rtos/gcc/pull/15 is merged.

PR #572, opened 2022-10-06
ci: Shorten distribution bundle path on Windows for testing

stephanosio:
This commit updates the CI workflow to shorten the distribution bundle path on Windows for testing because the full path can be too long for some architectures and may cause build failures.


stephanosio (commented):
Tested in https://github.com/zephyrproject-rtos/sdk-ng-testing/actions/runs/3194829602

PR #573, opened 2022-10-06
toolchain: Add AMD/Xilinx Microblaze support

asayin-xlnx:
First attempt to see if I can make the SDK build microblaze gcc too.

PR #578, opened 2022-10-25
toolchain: Add AMD/Xilinx Microblaze support

asayin-xlnx:
Attempt no.2. Previous attempt here: #573 

This new draft points to asayin-xlnx/gcc which has 1 extra commit which attempts to remove the buggy `moddi3` implementation.

https://github.com/zephyrproject-rtos/gcc/commit/c6eab4bac4d4aee7cd67f061a2f2c4d3bad8841d


stephanosio:
Rebased since the PR branch was outdated.

alpsayin:
Update: `moddi3` removal patch (https://github.com/zephyrproject-rtos/gcc/pull/16/commits/0d5c67b057b6192ee11e2322ae5275bf911d63ad) fixed the issue.
I'm now starting an internal review for the microblaze port (this might take a while, as this is more of a background task really). 
How would you like to proceed? 
  - Hold onto this PR until I can send a PR for zephyr? 
  - Hold onto this PR until I a PR for zephyr is actually gets merged? 
   - OR Merge this so it'd be easier to go back and forth between patches for zephyr?

stephanosio:
> How would you like to proceed?

It is really up to you. If you plan to upstream the microblaze arch port to Zephyr soon, then it would make sense to merge this now and provide a more fundamental fix later (assuming that is necessary; if not, we can just keep this too).

alpsayin:
Ok then, let's merge! 
As far issues go `moddi3` was I'd say a critical one. Thanks again for catching that one. The other "big" fail is the C++ exceptions which I made skip that particular test (honestly, I'm even surprised rest of the libcxx passed). I also flagged both issues in gcc bugzilla since they were also unpatched there.
I cant list in detail but unit tests generally put a good amount of stress to the port itself that I was able to catch many of my own bugs. And I say stress because I caught port bugs in fairly unrelated tests. For the same reason I'm hoping that toolchain will not show major issues.

alpsayin:
One quick question before anything else; we (I?) should probably change this last commit message, right?

stephanosio:
https://github.com/zephyrproject-rtos/gcc/pull/16 has been merged. Please update the submodule reference.

stephanosio:
@asayin-xlnx Is this ready to go in? If so, can you please take it out of draft status?

stephanosio (approved):

PR #580, opened 2022-10-25
A few SDK fixes from picolibc work

keith-packard:
These are generally applicable fixes which address bugs in the current SDK:

 1. Make sure we don't expose compiler TLS support for targets where Zephyr doesn't (yet) have TLS support
 2. Add Zephyr TLS register definition on arc so that code in the SDK that *does* use TLS works correctly.
 3. Apply an aarch64/gcc specific hack that Zephyr uses to the SDK.


dcpleung (approved):

configs/arc-zephyr-elf.config
```diff
@@ -4,7 +4,7 @@ CT_EXPERIMENTAL=y
 # CT_LOG_PROGRESS_BAR is not set
 CT_ARCH_ARC=y
 CT_MULTILIB=y
-CT_TARGET_CFLAGS="-mno-sdata"
+CT_TARGET_CFLAGS="-mno-sdata -mtp-regno=26"
```
abrodkin, line 7: @claziss could you please comment on the register used for bare-metal TLS implementation on both ARCv2 & ARCv3? To me it looks like it's `r25` at least on ARCv2.

evgeniy-paltsev: The register number itself is fine (I guess), we use r26 in zephyr itself: https://github.com/zephyrproject-rtos/zephyr/blob/9331591ab1f6a59513ded32d877f62dbf4526b9e/arch/arc/CMakeLists.txt#L15-L20I

But I'm wondering if it is OK to `-mtp-regno=26` to the toolchain itself, given that we don't always use TLS.

stephanosio: @keith-packard should we specify `-mtp-regno=26` for picolibc only? we can make picolibc depend on TLS.

claziss: Hi, TLS is default off for baremetal apps as they need to provide an environment for make it to work. Now, one can enable it or, if it is enabled, change the thread pointer (TP) register number using `mtp-regno` option. By default TP is set to r25, and it should work for other values as well, but I haven't test that.

keith-packard: > @keith-packard should we specify `-mtp-regno=26` for picolibc only? we can make picolibc depend on TLS.

Hrm. Does the compiler reserve this register when specified like this so that other code doesn't use it? If so, then, yes, we need to specify this for the toolchain so that no other code built by the toolchain conflicts with potential TLS usage by Zephyr. 

If the compiler doesn't reserve register 26, then does this parameter have any effect on code that doesn't use __thread variables? If not, then adding it unconditionally should not cause any issues.

In either case, I think we want to have the toolchain built with this parameter so that everything agrees on the TLS ABI.

keith-packard: Maybe we need to actually configure the compiler to have a defined TLS register and not hope that every application passes the right parameter?

keith-packard: ah, looking at the GCC source makes this all clear. When built for Linux targets, the default tp regno is 25. When build for elf targets, the default is -1, or unspecified. Take a look in gcc/config/arc/{elf.h,linux.h} for details.

Is there some reason zephyr doesn't follow the Linux convention and use register 25 for this? Shouldn't we patch gcc to set `TARGET_ARC_TP_REGNO_DEFAULT` instead of expecting everyone to agree to pass the correct value on the gcc command line?

stephanosio: @abrodkin @evgeniy-paltsev any comments on the above feedback?

stephanosio: > We've discussed this - let's use r25 in arc-zephyr-elf as tls to allign with other ARCv2 toolchains.

@evgeniy-paltsev sure, but can you please create a separate PR for this to `sdk-ng` and `zephyr` since it is quite a major change and not really something for the SDK 0.15.2?

> I think all code linked together would need to have this specified to ensure it doesn't smash the TLS register -- I'd assume the compiler would be free to use it otherwise?

That would come with some performance implications when TLS is not enabled; but, I assume such a performance impact would be small enough.

Given that it probably is not going to be worth building separate multilib variants for non-TLS and TLS, sure.

keith-packard: > Given that it probably is not going to be worth building separate multilib variants for non-TLS and TLS, sure.

This raises a good point -- we should stick the -mtp-regno parameter into the multilib configuration -- that way a conflicting value will cause link failures instead of silent corruption. And, any pieces of the toolchain using the multilib configs (like newlib, picolibc, libstdc++) will automatically get the correct value here. That seems like a good way to keep the toolchain os-independent and ensure that Zephyr uses will work correctly.

keith-packard: and, of course, would allow us to build multilib configurations that use r25 or r26 even, if the Arc team wants to transition.

keith-packard: I'm giving that a try to see how bad it gets. "only" multiplies the library builds for the toolchain by a factor of three ...

stephanosio:
DNM until https://github.com/zephyrproject-rtos/gcc/pull/17 is resolved.

stephanosio:
To be merged as part of:
* #587 
* #588 
* #589

PR #585, opened 2022-11-04
meta-zephyr-sdk: Use glibc 2.27

stephanosio:
The current Poky distribution includes the glibc 2.32, which is too recent for acceptable backwards compatibility.

This commit adds the glibc 2.27 recipe to the meta-zephyr-sdk in order to force the host tool components to target the glibc version 2.27.

The glibc version 2.27 is the version of glibc included in Ubuntu 18.04, which is the minimum version of Ubuntu that Zephyr SDK aims to be compatible with.


keith-packard:
Not sure about the implementation, but the goal seems correct.

stephanosio:
> Not sure about the implementation, but the goal seems correct.

Yea, I am just trying to see if I can use this ugly hack to override the glibc recipe from Yocto, but it looks like there have been too many changes since the last Yocto version using glibc 2.27 for this to be even feasible. It looks like I will have to downgrade the Yocto version.

PR #588, opened 2022-11-06
arc: Use register 26 for TLS storage.

keith-packard:
Include gcc version with change that specifies default tp-regno for arc target so that the whole toolchain will agree.

Signed-off-by: Keith Packard <keithp@keithp.com>

PR #595, opened 2022-11-17
arc: Add -mtp-regno=26 to CFLAGS

stephanosio:
Make sure everything is built with the correct TLS configuration.

---

From https://github.com/zephyrproject-rtos/zephyr/pull/51910#issuecomment-1310191743:

> Drop https://github.com/zephyrproject-rtos/gcc/pull/17 PR. As we can do it without changes in GCC by simple toolchain configuration tweak - let's do it that way. We prefer to not have extra patches for GCC for ARC in Zephyr GCC fork which we don't upstream to GCC.

As per the above, this PR takes the previous patch adding `-mtp-regno=26` to the ARC toolchain configurations from https://github.com/zephyrproject-rtos/sdk-ng/pull/580 so that we have something that works for the SDK 0.15.2 release.

Note that this is a temporary solution until #588 or a more complete solution described in https://github.com/zephyrproject-rtos/zephyr/pull/51910#issuecomment-1310191743 is implemented.


stephanosio:
FYI @keith-packard @evgeniy-paltsev

PR #602, opened 2022-11-24
Add picolibc

keith-packard:
This adds picolibc as a module and builds it for all SDK architectures. To make the resulting files smaller than 2GB, the compression algorithm for Windows is switched to 7z while the Linux files use .tar.xz.

Picolibc is pulled at a point which has not yet been released upstream, but I expect to do that soon. Do you want to use a released version of picolibc for the Zephyr SDK?


stephanosio:
> Picolibc is pulled at a point which has not yet been released upstream, but I expect to do that soon. Do you want to use a released version of picolibc for the Zephyr SDK?

Nope, that is fine as long as it works. We will have a month or so until the SDK 0.16.0 release, so we can update the picolibc submodule during that time if any further changes are necessary.

stephanosio (approved):

PR #626, opened 2023-02-09
kconfig: Signal newlib support in Kconfig too

keith-packard:
This lets Kconfig stanzas detect newlib support in the toolchain

Working on zephyr PR https://github.com/zephyrproject-rtos/zephyr/pull/54637

### DIFF
--- a/cmake/zephyr/Kconfig
+++ b/cmake/zephyr/Kconfig
@@ -7,3 +7,9 @@ config TOOLCHAIN_ZEPHYR_0_16
 config TOOLCHAIN_ZEPHYR_SUPPORTS_THREAD_LOCAL_STORAGE
 	def_bool y
 	select TOOLCHAIN_SUPPORTS_THREAD_LOCAL_STORAGE
+
+config NEWLIB_SUPPORTED
+	def_bool y
+	depends on "$(ZEPHYR_TOOLCHAIN_VARIANT)" = "zephyr"
+	help
+	  Zephyr always supports newlib for C and C++ development.


cmake/zephyr/Kconfig
```diff
@@ -7,3 +7,9 @@ config TOOLCHAIN_ZEPHYR_0_16
 config TOOLCHAIN_ZEPHYR_SUPPORTS_THREAD_LOCAL_STORAGE
 	def_bool y
 	select TOOLCHAIN_SUPPORTS_THREAD_LOCAL_STORAGE
+
+config NEWLIB_SUPPORTED
```
galak, line 11: I don't think a generic option should be in here.  For example how will the arm-gnu-embedded toolchain enable `NEWLIB_SUPPORTED`.  This should be in Zephyr proper and have something like:

```
config TOOLCHAIN_ZEPHYR_SUPPORTS_NEWLIB`
  def_bool y
  select NEWLIB_SUPPORTED
```

galak (changes requested):
Put generic Kconfig options like this in Zephyr proper, and just select them here.

cmake/zephyr/Kconfig
galak: or add `select NEWLIB_SUPPORTED` under `config TOOLCHAIN_ZEPHYR_0_16`

keith-packard: I've already got a proposed patch which does that (no need for the version check, Zephyr SDK has always supported newlib). I was just following the suggestion about where similar mechanism for Picolibc should live, and that will live in sdk-ng. Happy to have it in Zephyr instead; it's a lot easier to manage there anyways.

stephanosio (changes requested):
See https://github.com/zephyrproject-rtos/zephyr/pull/54637#discussion_r1102742570

PR #632, opened 2023-03-03
cmake: improve text description on minimum compatible versions

tejlmand:
During review of https://github.com/zephyrproject-rtos/zephyr/pull/53833 it became clear that existing description regarding minimum compatible version was unclear. Also the variable name
`ZEPHYR_MINIMUM_COMPATIBLE_VERSION` could be mistaken for referring to Zephyr itself, and not the Zephyr SDK.

Improve the description to be more precise that we are referring to Zephyr SDK versions. Update `ZEPHYR_MINIMUM_COMPATIBLE_VERSION` to `ZEPHYR_SDK_MINIMUM_COMPATIBLE_VERSION`.

### DIFF
--- a/cmake/Zephyr-sdkConfigVersion.cmake
+++ b/cmake/Zephyr-sdkConfigVersion.cmake
@@ -14,10 +14,15 @@ else()
   # Currently, this Zephyr SDK is expected to work with any Zephyr project
   # requiring this version or any older version.
   #
-  # In case this version is no longer backward compatible then this is the place
-  # to test, for example as
-  # set(ZEPHYR_MINIMUM_COMPATIBLE_VERSION 0.11)
-  # if(PACKAGE_FIND_VERSION VERSION_LESS ZEPHYR_MINIMUM_COMPATIBLE_VERSION)
+  # In case this version of Zephyr SDK is no longer backwards compatible with
+  # previous versions of Zephyr SDK, this is the place to test if the caller
+  # requested an older (incompatible) revision.
+  # For example, imagine caller requests Zephyr SDK v0.10, and this Zephyr SDK
+  # is revision v0.11, then the below snippet can be activated to ensure that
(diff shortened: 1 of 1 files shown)


tejlmand:
@nordicjm please take a look. I cannot add you as reviewer but your feedback would be appreciated.

nordicjm (approved):

cmake/Zephyr-sdkConfigVersion.cmake
```diff
@@ -14,10 +14,15 @@ else()
   # Currently, this Zephyr SDK is expected to work with any Zephyr project
   # requiring this version or any older version.
   #
-  # In case this version is no longer backward compatible then this is the place
-  # to test, for example as
-  # set(ZEPHYR_MINIMUM_COMPATIBLE_VERSION 0.11)
-  # if(PACKAGE_FIND_VERSION VERSION_LESS ZEPHYR_MINIMUM_COMPATIBLE_VERSION)
+  # In case this version of Zephyr SDK is no longer backward compatible previous
+  # versions of Zephyr SDK this is the place to test if caller requested an
+  # older (in-compatible) revision.
```
dkalowsk, line 19: There are a few words missing here making it a little unclear what is being communicated.  Below is my best guess towards understanding the intent:

```
In case this verison of Zephyr SDK is no longer backwards compatible with
previous versions of Zephyr SDK, this is the place to test if the caller requested
an older (in-compatible) revision.
```

keith-packard: incompatible isn't usually hypenated

tejlmand: fixed, thanks for the suggestions.

dkalowsk (approved):

stephanosio (approved):

PR #635, opened 2023-03-07
ci: Disable testing of xtensa-espressif_esp32s3_zephyr-elf toolchain

stephanosio:
This commit disables the testing of the
`xtensa-espressif_esp32s3_zephyr-elf` toolchain because the `esp32s3_devkitc` board has not been upstreamed yet.

Revert this commit when the `esp32s3_devkitc` board is upstreamed.


stephanosio:
@sylvioalves I see that there currently is `esp32s3_devkitm` board in the upstream Zephyr. Any reason we have to use devkit**c** instead of devkit**m** here in the SDK CI?

sylvioalves:
@stephanosio, sorry, should have been `esp32s32_devkitm`

Issue #633, opened 2023-03-03
add TERM support for `tmux-256color` and `screen-256color`


mtfurlan:
Zephyr currently has a [hardcoded list of terminal types it supports](https://github.com/zephyrproject-rtos/sdk-ng/issues/576#issuecomment-1276447224): 
> `linux,xterm,xterm-color,xterm-256color,vt100`

This makes life hard if you use tmux, because forcing tmux to use `xterm-256color` instead of `screen-256color` or `tmux-256color` means the Home and End keys don't work properly, and likely other stuff as I only tested for a day or so.

After some discussion with @stephanosio on discord, the best solution seems to be just to add `screen-256color` and `tmux-256color` to the zephyr terminfo database because tmux is a very widely used program.

PR #642, opened 2023-03-09
configs: common: Add tmux support for ncurses

stephanosio:
This commit updates the common toolchain configurations to explicitly set the list of supported terminal types for the ncurses library and adds `screen`, `screen-256color`, `tmux`, `tmux-256color` terminal type support.

These terminal types are required in order to properly handle user inputs (e.g. for GDB shell) inside tmux.

---

Closes https://github.com/zephyrproject-rtos/sdk-ng/issues/633

### DIFF
--- a/configs/common.config
+++ b/configs/common.config
@@ -1,5 +1,8 @@
 # Common crosstool-ng configurations for all toolchain variants
 
+# Ncurses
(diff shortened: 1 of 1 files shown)


stephanosio (commented):
Verified that key inputs (e.g. backspace) to GDB work properly inside tmux.

PR #643, opened 2023-03-09
ci: Grant write permission before toolchain output directory cleanup

stephanosio:
This commit updates the CI workflow to grant write permission to the toolchain build output directory and its sub-directories before attempting to clean up unneeded files.

This is necessary because crosstool-ng may create output directories without write permission (e.g. when building on macOS).


stephanosio (commented):
Manually tested in the CI runner.

Merging as a CI hotfix.

PR #645, opened 2023-03-13
VERSION: Bump to 0.16.0

stephanosio:
Release Zephyr SDK 0.16.0


keith-packard:
Alp Sayin and I have been hacking on microblazeel, trying to get picolibc tests running. It looks like atomics aren't working yet, so there may be some gcc issue (we're using -rc1 for testing). We also have fixes for picolibc on microblazeel that aren't finished yet.

keith-packard:
Looks like at least this patch is missing upstream: https://marc.info/?l=microblaze-uclinux&m=141825239401679&q=mbox

stephanosio:
> Alp Sayin and I have been hacking on microblazeel, trying to get picolibc tests running. It looks like atomics aren't working yet, so there may be some gcc issue (we're using -rc1 for testing). We also have fixes for picolibc on microblazeel that aren't finished yet.

microblazeel support is experimental and only for initial bring-up and testing, so it should not block the release -- any issues with it can be fixed in a later release.

@alpsayin If there are any microblazeel toolchain issues that need to be fixed in the next SDK release, please file a bug in the sdk-ng repository.

alpsayin:
@stephanosio @keith-packard and I caught one more regarding atomic-ungetc just today. Maybe that.

alpsayin:
I should've read the prior message. Yes, will do with a bug report and a future PR.

dkalowsk (approved):
So far haven't had any issues on the arm builds for 0.16

keith-packard:
> microblazeel support is experimental and only for initial bring-up and testing, so it should not block the release -- any issues with it can be fixed in a later release.

sounds good. Do you want a PR for picolibc to have the SDK match what the Zephyr module is using? The only big difference is the addition of wchar_t support so that we can expose a 'full' libstdc++ API.

stephanosio:
> sounds good. Do you want a PR for picolibc to have the SDK match what the Zephyr module is using? The only big difference is the addition of wchar_t support so that we can expose a 'full' libstdc++ API.

I think it would be better to leave that as an enhancement for a future release since wide character support is not a critical feature for Zephyr (or embedded systems in general) and introducing that change at this point can potentially lead to regressions.

keith-packard:
> I think it would be better to leave that as an enhancement for a future release since wide character support is not a critical feature for Zephyr (or embedded systems in general) and introducing that change at this point can potentially lead to regressions.

Sounds fine. Seems a bit unexpected for the Zephyr module to be out of sync from the SDK, but I guess that's actually not at all unlikely. We can wait to update the SDK until the microblaze bits actually work.

keith-packard (approved):
I've been building and testing with this branch without issue for quite a while.

PR #658, opened 2023-04-10
ci: Disable testing of xtensa-intel_apl_adsp_zephyr-elf

stephanosio:
This commit updates the CI workflow to disable testing of the `xtensa-intel_apl_adsp_zephyr-elf` toolchain because the `intel_adsp_cavs15` board was removed from the upstream Zephyr and there currently is no upstream board that uses this toolchain.


stephanosio:
https://github.com/zephyrproject-rtos/zephyr/pull/56588

PR #664, opened 2023-04-19
xtensa: intel_adsp: toolchain shuffling due to board removals

dcpleung:
This removes both `intel_apl_adsp` and `intel_s1000` since there are no boards using these anymore. Also adds `intel_tgl_adsp` so the board `intel_adsp_cavs25` can use a proper toolchain.


dcpleung:
This is needed for https://github.com/zephyrproject-rtos/zephyr/pull/55840

stephanosio (approved):

PR #666, opened 2023-04-22
binutils: Pull in ARM MVE VCVTNE misinterpretation fix

### LINKED ISSUE #665: gas: Misinterpretation of the the ARM MVE VCVTNE instruction
From https://github.com/bminor/binutils-gdb/commit/656412a7a4da0aef43ead1ea976ba0235a7ec30a:

Previously we had experienced issues with assembling a "VCVTNE" instruction
in the presence of the MVE architecture extension, because it could be
interpreted both as:

* The base instruction VCVT + NE for IT predication when inside an IT block.
* The MVE instruction VCVTN + E in the Else of a VPT block.

Given a C reproducer of:
```
int test_function(float value)
{
  int ret_val = 10;
  if (value != 0.0)
  {
    ret_val = (int) value;
  }
  return ret_val;
}
```
GCC generates a VCVTNE instruction based on the `truncsisf2_vfp`
pattern, which will look like:
`vcvtne.s32.f32 s-reg, s-reg`
This still triggers an error due to being misidentified as "vcvtn+e"
Similar errors were found with other type combinations and instruction
patterns (these have all been added to the testing of this patch).

This class of errors was previously worked around by:
https://sourceware.org/pipermail/binutils/2020-August/112728.html
which addressed this by looking at the operand types, however,
that isn't adequate to cover all the extra cases that have been
found.  Instead, we add some special-casing logic earlier when
the instructions are parsed that is conditional on whether we are
in a VPT block or not, when the instruction is parsed.

stephanosio:
This commit pulls in the binutils/gas patch that fixes the misinterpretation of the MVE `VCVTN` instruction as the base `VCVT` instruction.

---

GDB PR: https://github.com/zephyrproject-rtos/binutils-gdb/pull/8

Fixes #665


stephanosio:
DNM until https://github.com/zephyrproject-rtos/binutils-gdb/pull/8 is merged.

stephanosio (commented):
Tested and verified that `arm-zephyr-eabi` toolchain can now successfully compile the following test vector:

```
int test_function(float value)
{
  int ret_val = 10;
  if (value != 0.0)
  {
    ret_val = (int) value;
  }
  return ret_val;
}
```

PR #677, opened 2023-05-26
cmake: arc: provide helper tools for arcmwdt toolchain

evgeniy-paltsev:
As of today MWDT toolchain in case of Zephyr may use few tools from GNU toolchain - GNU compiler for DTS preprocessing and objcopy for section renaming.

Currently we were trying to find any GNU compiler & objcopy which start to cause compatibility issues for new targets - i.e. not every objcopy knows new ARC targets.

Alternative approach is to use ARC GNU tools from Zephyr SDK as we require it when building with MWDT for other tools like DTC anyway.

This change simplify usage ARC GNU tools from Zephyr SDK for ARC MWDT toolchain.

This intended to be use with proper changes on the Zephyr side: https://github.com/zephyrproject-rtos/zephyr/pull/58329


abrodkin (approved):

tejlmand (changes requested):
not fond of adding extra hard couplings between Zephyr and the internal file structure of Zephyr SDK as proposed with this file:
https://github.com/zephyrproject-rtos/sdk-ng/blob/ea6236ba9444055a7dd4e48c985c51557ebd302c/cmake/zephyr/arcmwdt-helper.cmake

Also this introduces specific ARCMWDT helper symbols which is ARCMWDT knowledge inside Zephyr SDK.

I much prefer we keep Zephyr SDK itself unrelated to arcmwdt.

I made a proposal on how integration with existing Zephyr SDK releasses can be done without requiring this PR.

That said, this PR, together with https://github.com/zephyrproject-rtos/zephyr/pull/58329 has revealed that Zephyr SDK does need an update to its CMake integration to make it more re-usable / integration friendly, so thanks for revealing that 👍

evgeniy-paltsev:
As the https://github.com/zephyrproject-rtos/zephyr/pull/58329 was changed according to Torsten's proposal - we don't need this change for SDK anymore.

PR #187, opened 2020-02-10
Xtensa targets from SOF

lgirdwood:
This PR is one of the first steps in adding support for SOF as a Zephyr application and takes all the Xtensa toolchain recipes from the supported Intel and NXP SOF platforms and imports them as Zephyr SDK toolchains.

This PR allows building for Intel Baytrail, Cherrytrail, Braswell, Broadwell and NXP IMX8 and IMX8M audio DSPs. The other SOF target DSPs are already supported using the S1000 and APL toolchains.

PR #694, opened 2023-08-07
xtensa: Add mt8195_adsp toolchain

andyross:
Yet another Xtensa SOC for the SDK.  Mostly the same as the other LX cores, but different enough to not be able to share code. (No MULUH instruction on this model, which breaks on compiler-generated 64 bit multiplies).

[Submitting this blindly to see if a usable SDK pops out magically, as per the docs.  Hopefully there's help available when something breaks.  Isn't there a Dockerfile or something I can use to do a build somewhere?]


.github/workflows/ci.yml
```diff
@@ -1531,6 +1535,9 @@ jobs:
             xtensa-intel_tgl_adsp_zephyr-elf)
               PLATFORM_ARGS+="-p intel_adsp_cavs25 "
               ;;
+            xtensa-mt8195_adsp_zephyr-elf)
+              PLATFORM_ARGS+="-p intel_adsp_cavs25 "
```
dcpleung, line 1539: Will need to update this after the board config has been merged.

andyross: Oops, yeah, that's a cut/paste error for sure.  Will fix.

dbaluta:
@andyross This looks good to me. 

FYI: https://github.com/zephyrproject-rtos/sdk-ng/pull/187/commits here is how @lgirdwood did the changes for Inte/NXP.

.github/workflows/ci.yml
```diff
@@ -163,6 +164,7 @@ jobs:
             xtensa-espressif_esp32s3_zephyr-elf)  build_target_xtensa_espressif_esp32s3_zephyr_elf="y";;
             xtensa-intel_ace15_mtpm_zephyr-elf)   build_target_xtensa_intel_ace15_mtpm_zephyr_elf="y";;
             xtensa-intel_tgl_adsp_zephyr-elf)     build_target_xtensa_intel_tgl_adsp_zephyr_elf="y";;
+            xtensa-mt8195_adsp_zephyr-elf)     build_target_xtensa_mt8195_adsp_zephyr_elf="y";;
```
sylvioalves, line 167: For the sake of alignment, would you add a line break here? `build_target....`

dbaluta:
@andyross  I'm also interested about the toolchain automagically being built by the CI.

Anyone can shed some light on this? :)

dbaluta (approved):

nashif:
> But the new toolchain isn't there. It only built the old ones. I'm pretty sure it was added to all the places needed (basically I cloned the settings for intel_tgl_adsp, everywhere it's mentioned there's a line for mtk8195_adsp). Is the issue maybe that CI runs the workflow from the release and not the code in the PR? How do I fix this?

this is because of the action mode (pull_request_target). security blah blah

lgirdwood (approved):

dbaluta:
@andyross  this needs rebase.

andyross:
Rebased & fixed up, should hopefully work.

FWIW: I noticed while mucking with this a while back that there appears to be (vestigial?) support for multi-target gcc xtensa builds upstream.  Has anyone every looked at this?  Basically there's an "xtensa-dynconfig.c" file in upstream binutils and gcc, and it appears to look at a "XTENSA_GNU_CONFIG" environment variable, which it appears to interpret as a shared library that it dlopens to get the various tables normally built out of the overlay.

Has anyone else every noticed this or tried it?   This would save HUGE amounts of space in the SDK if we could make it work.  It looks semi-abandoned, but not complicated.

andyross:
(FWIW: also need to get the mt8195 HAL additions updated too, I think there were comments.  But the board PR itself is currently buried and horribly entangled with the call0 work, so that's going to take a bit to unravel)

overlays/xtensa_mtk_mt8195/gdb/gdb/xtensa-config.c
```diff
@@ -0,0 +1,504 @@
...
+            0,0,&xtensa_mask38,0,0,0)
+  XTREG_END
+};
+
+
+
+#ifdef XTENSA_CONFIG_INSTANTIATE
+XTENSA_CONFIG_INSTANTIATE(rmap,16)
+#endif
+
```
dcpleung, line 504: Need to add:
```
xtensa_gdbarch_tdep xtensa_tdep (rmap);
```
or else it won't build.

dcpleung (changes requested):

overlays/xtensa_mtk_mt8195/gdb/gdb/xtensa-config.c
andyross: Sorry, I'm lost on that requirement.  Where exactly do I put that and why?  :)

This is just the upstream file from the Cadence SDK (with some path munging via that make-overlay.sh script at foss-xtensa).  Is this something they've forgotten or that we need to maintain and didn't document?  Or am I looking at the wrong place?

andyross: Ah, OK.  Skew vs. upstream gdb APIs.  Found the other commits in the tree.

We really need to write a "overlay importer" layer or something.  To add to the list with "make dynamic configs work" I guess.  Managing this is sort of a mess for someone doing it the first time, and we're juggling a ton of these toolchains these days.

.github/workflows/ci.yml
```diff
@@ -1546,6 +1550,9 @@ jobs:
             xtensa-intel_tgl_adsp_zephyr-elf)
               PLATFORM_ARGS+="-p intel_adsp_cavs25 "
               ;;
+            xtensa-mt8195_adsp_zephyr-elf)
+              PLATFORM_ARGS+="-p mt8195_adsp "
```
nashif, line 1554: Do we have a PR introducing this? We need to have this ready somehow and merge it or put it in the sdk branch to get this test to build, otherwise CI will fail.

andyross: PR for the board is at https://github.com/zephyrproject-rtos/zephyr/pull/61197 but I wasn't pushing it as there are still some review comments and IMHO it makes more sense to wait for the call0 work to land first as then it can launch with working userspace (it's got an MPU).

Isn't that a chicken and egg problem?  I assumed we needed the SDK patch before the board could merge too?  Maybe I should remove these lines and merge an update later?

dcpleung (approved):

.github/workflows/ci.yml
nashif: it is a chicken and egg problem, but when we test the SDK in CI, we pull this from a branch (topic-sdk-dev), so we can put this PR for building purposes and while we wait for it to get merged in zephyr main.

PR #700, opened 2023-08-30
xtensa: mtk_mt8195_adsp: unify naming to mtk_mt8195_adsp

dcpleung:
This unifies various naming of mt8195 to mtk_mt8195_adsp which is of scheme xtensa-(vendor)_(soc name)-elf (except the QEMU cores). This renames the overlay directory, and updating both the config file and github workflows to match.


andyross (approved):
Was this the source of the build failure?  Something was inconsistent or assuming there were three components?  No complaints here about the name beyond verbosity.  I guess I must have been thinking along the lines of: "mt" in "mt8195" already implies the vendor.  But sure.

dcpleung:
The naming mismatch between the middle part of the config file name (the part between `xtensa-` and `-elf`), and `CT_TARGET_VENDOR` results in a different output directory being created during the install phase of tools (think make install). So the script couldn't find the output directory and cannot further process the artifacts (e.g. stripping the binaries and remove the duplicate sysroot).

nashif (approved):

PR #701, opened 2023-08-31
xtensa: enable multilib for dc233c and sample_controller

dcpleung:
This enables multilib support for building GCC. This is needed to support both CALL0 and Windowed ABIs.


andyross (approved):
Magnificent

dcpleung:
If CI is building this correctly, we should be able to grab the artifacts in https://github.com/zephyrproject-rtos/sdk-ng/actions/runs/6043012028

nashif:
@dcpleung  why is this still a fraft? Can you rebase?

dcpleung:
This is needed for CALL0 API support. Since we still have no viable CALL0 code to test, there is not much point in merging it now. That's why it's in draft until we have something to test with.

PR #702, opened 2023-09-06
xtensa: dc233c: disable thread local storage

dcpleung:
The main Zephyr arch code does not support thread local storage yet. So set CT_CC_GCC_CONFIG_TLS to NO so picolibc would not be built with TLS support.


dcpleung:
not needed anymore.

PR #731, opened 2024-01-25
ci: Disable AArch64 Linux SDK build

stephanosio:
This commit temporarily disables AArch64 Linux Zephyr SDK build because the AArch64 CI runner AWS spot instances are too frequently shut down and unable to complete the full build cycle.

Re-enable this when the self-hosted zephyr-runner v2 is deployed and available for production use.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,7 +187,7 @@ jobs:
 
         if [ "${build_host_all}" == "y" ]; then
           build_host_linux_x86_64="y"
-          build_host_linux_aarch64="y"
+          # build_host_linux_aarch64="y"
           build_host_macos_x86_64="y"
           build_host_macos_aarch64="y"
           build_host_windows_x86_64="y"


rexut:
> This commit temporarily disables AArch64 Linux Zephyr SDK build because the AArch64 CI runner AWS spot instances are too frequently shut down and unable to complete the full build cycle.

That are bad news. Our hardware-in-the-loop infrastructure for different product lines, based on Raspberry Pi 4, needs the AArch64 binaries.

> 
> Re-enable this when the self-hosted zephyr-runner v2 is deployed and available for production use.

Is there a termination date for this new rollout? In this new situation, I have to calculate the risks and consequences for our production, and possibly build everything myself. This will take time and it would be foolish to do it if the AArch64 packages are already available before that.

stephanosio:
> Is there a termination date for this new rollout?

I am hoping to make a follow-up SDK release with AArch64 binaries before the Zephyr 3.6 release date (planned to be 2024/02/23) -- I cannot make any promises though.

If not, I can say with high certainty that it will be no later than mid-March.

PR #647, opened 2023-03-13
gcc: Pull in Microblaze gcc atomic cas return value fix

alpsayin:
This commit pulls in the fix for the Microblaze gcc atomic compare-and-swap issue where atomic cas returns the inverse boolean value of what it should. https://github.com/zephyrproject-rtos/gcc/18

Fixes https://github.com/zephyrproject-rtos/sdk-ng/issues/646

PR #742, opened 2024-03-15
gcc: Add support for Arm Cortex-M85

thaoluonguw:
This adding supports for building GCC with Arm Cortex-M85.

Need: https://github.com/zephyrproject-rtos/gcc/pull/28


.gitmodules
```diff
@@ -9,7 +9,7 @@
 	url = https://github.com/zephyrproject-rtos/newlib-cygwin.git
 [submodule "gcc"]
 	path = gcc
-	url = https://github.com/zephyrproject-rtos/gcc.git
+	url = https://github.com/thaoluonguw/gcc.git
```
tejlmand, line 12: we should get this merged and thus picked automatically: https://github.com/zephyrproject-rtos/gcc/pull/28
FYI @stephanosio

tejlmand (changes requested):

.gitmodules
thaoluonguw: @tejlmand, https://github.com/zephyrproject-rtos/gcc/pull/28 is merged. I updated submodule to reference the merged commit in gcc. Please help me review for it. Thank you.

stephanosio (approved):

stephanosio:
Note that this advances the GCC ref past the already merged GCC patches for #647 and #735 -- this is not a problem because those patches are approved/accepted and some parts of it are no-op until activated by the corresponding sdk-ng PRs.

PR #745, opened 2024-04-11
ci: Increase macOS workspace volume size to 150GB

stephanosio:
This commit increases the macOS runner workspace volume size from 70GB to 150GB because 70GB is insufficient for the toolchain builds for the architectures with many multi-lib variants to successfully complete.

Note that the size of 150GB was chosen to match the workspace volume size available on the Linux zephyr-runner v2 instances.


stephanosio (commented):
Tested in https://github.com/zephyrproject-rtos/sdk-ng/actions/runs/8647815150

PR #750, opened 2024-04-29
ci: Update test platform names for HWMv2

stephanosio:
Update test platform names for HWMv2 after the `collab-sdk-dev` branch fast-forward.

While at it, also update the test target toolchain list to include all currently available toolchains.


stephanosio (commented):
Test failure is expected since the workflow runs in the context of the base branch and the changes in this PR are not reflected.

Merging with a failure.

Issue #760, opened 2024-05-21
undefined reference to `_impure_ptr' when building Zephyr C++ samples on non-Linux hosts


stephanosio:
Zephyr C++ samples and tests fail to link with "undefined reference to `_impure_ptr'" on macOS and Windows hosts using the respective host toolchains.

For example, building `samples/cpp/hello_world` for `qemu_arc_hs` on a Windows host results in the following failure:

```
[99/104] Linking CXX executable zephyr\zephyr_pre0.elf

FAILED: zephyr/zephyr_pre0.elf zephyr/zephyr_pre0.map C:/a/sdk-ng/sdk-ng/test/twister-out/twister_links/test_442/zephyr/zephyr_pre0.map 

[...]

PR #761, opened 2024-05-23
ci: Exclude C++ sample and test builds on non-Linux hosts

stephanosio:
This commit updates the CI workflow such that Zephyr C++ samples and tests are only built on Linux hosts because there is a known issue causing the following link error using non-Linux toolchains:

  undefined reference to `_impure_ptr'

Revert this when the GitHub issue #760 is fixed.


stephanosio (commented):
Tested in https://github.com/zephyrproject-rtos/sdk-ng/actions/runs/9177283395

PR #770, opened 2024-06-20
gcc: Pull Zero/Sign extends for CMSE security

stephanosio:
This commit pulls in the GCC patch that adds Zero/Sign extends for the CMSE secure mode transitions to address the CVE-2024-0151.

---

GCC PR: https://github.com/zephyrproject-rtos/gcc/pull/32


stephanosio:
DNM until https://github.com/zephyrproject-rtos/gcc/pull/32 is merged and this PR is updated to reference the merged commit.

PR #773, opened 2024-08-08
minichlink: create a recipie to build minichlink for the ch32v003

nzmichaelh:
minichlink is a self contained flasher, debug stub, and terminal for the CH32V003.

If this looks good then I'll send the Makefile patch upstream.


nzmichaelh:
Morning @stephanosio it looks like there's a generic nodejs / glibc version incompatibility in the build scripts. Any ideas?

stephanosio:
> Morning @stephanosio it looks like there's a generic nodejs / glibc version incompatibility in the build scripts. Any ideas?

Rebased to pull in the fixes.

Issue #764, opened 2024-05-28
Migrate Linux CI environment to Debian 10


stephanosio:
The current Zephyr SDK Linux CI build environment is based on Ubuntu 18.04 in order to ensure that the Linux Zephyr SDK binaries remain compatible with Ubuntu 18.04 and other distro releases of the same era (glibc 2.27).

This proposes to increase the minimum host "compatibility level" of the Linux Zephyr SDK binaries to Debian 10 (glibc 2.28) because:

1. Ubuntu 18.04 is over 6 years old at this point.
2. Ubuntu 18.04 has reached the end of "standard support" in April 2023.
3. Nobody is/should be using it on their development machine.

The following are the glibc versions of the major distro releases from the past 10 years:

Distro  | glibc
---|---
Debian 9 | 2.24
Debian 10 | 2.28
Debian 11 | 2.31
Debian 12 | 2.36
RHEL 7 | 2.17
RHEL 8 | 2.28
RHEL 9 | 2.34
Ubuntu 18.04 | 2.27
Ubuntu 20.04 | 2.31
Ubuntu 22.04 | 2.35
Ubuntu 24.04 | 2.39

Targeting glibc 2.28 would ensure compatibility with Debian 10+, RHEL 8+ and Ubuntu 20.04+, which is a fairly reasonable range of distro releases.

stephanosio:
Also note that GitHub has deprecated Node.js 16 actions and will soon bump the minimum supported Node.js version to 20.

Node.js, since the version 18, has increased the minimum required glibc version for the prebuilt Linux binaries to 2.28 and, unless we plan to build Node.js ourselves and hack the GHA runner to use it, we are forced to upgrade to a host environment with glibc 2.28 or above.

PR #778, opened 2024-09-10
qemu: Update QEMU version to the tagged 9.1.0

teburd:
This release of qemu has important bug fixes that perhaps will fix a bug in xtensa emulation of mmu.


teburd:
No idea what I'm doing here, hopefully this does what I want

stephanosio (changes requested):
This will not work because https://github.com/zephyrproject-rtos/qemu does not have v9.1.0.

Also note that we carry quite a few local patches in the aforementioned fork -- they need to be cherry-picked into the v9.1.0 branch.

p.s. Regarding the CI failure in general, that is because of https://github.com/zephyrproject-rtos/sdk-ng/issues/764#issuecomment-2134379357 and I am looking into it.

nashif:
lets close this, work on 9.2.1 is happening elsewhere.

PR #792, opened 2024-09-19
xtensa: Add xtensa-amd_acp_6_0_adsp_zephyr-elf.

stephanosio:
This adds the config and overlay files to enable building the toolchain needed for the acp_6_0 board in Zephyr.


stephanosio:
@DINESHKUMARK1 FYI. You can push directly to the upstream `topic-amd_acp_6_0` if you need to make any further changes.

PR #799, opened 2024-09-24
ci: Do not remove artifacts when running twister

stephanosio:
This removes the the `-M` (runtime artifact cleanup) parameter from twister invocation because it causes instability when running on Windows and the artifacts for the passing tests are already cleaned up without it.


stephanosio (commented):
Tested in https://github.com/zephyrproject-rtos/sdk-ng/actions/runs/11011682163

PR #802, opened 2024-09-26
Set correct OPENOCD_DEFAULT_PATH for Windows

Manu3l0us:
On Windows platform, the OPENOCD_DEFAULT_PATH is not set in line with Zephyr documentation. Fixes zephyrproject-rtos/zephyr#79051.

### DIFF
--- a/cmake/zephyr/host-tools.cmake
+++ b/cmake/zephyr/host-tools.cmake
@@ -10,4 +10,8 @@ list(APPEND CMAKE_PREFIX_PATH ${HOST_TOOLS_HOME}/usr)
 
 # TODO: Use find_* somehow for these as well?
 set_ifndef(QEMU_BIOS            ${HOST_TOOLS_HOME}/usr/share/qemu)
-set_ifndef(OPENOCD_DEFAULT_PATH ${HOST_TOOLS_HOME}/usr/share/openocd/scripts)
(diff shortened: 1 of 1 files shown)


cmake/zephyr/host-tools.cmake
```diff
@@ -10,4 +10,8 @@ list(APPEND CMAKE_PREFIX_PATH ${HOST_TOOLS_HOME}/usr)
 
 # TODO: Use find_* somehow for these as well?
 set_ifndef(QEMU_BIOS            ${HOST_TOOLS_HOME}/usr/share/qemu)
-set_ifndef(OPENOCD_DEFAULT_PATH ${HOST_TOOLS_HOME}/usr/share/openocd/scripts)
+if (WIN32)
+    set_ifndef(OPENOCD_DEFAULT_PATH "C:/Program Files/OpenOCD/share/openocd/scripts")
```
stephanosio, line 14: Please use `ENV{ProgramFiles}`.

stephanosio (changes requested):

cmake/zephyr/host-tools.cmake
Manu3l0us: Done, thx.

stephanosio (approved):

PR #819, opened 2024-10-10
picolibc: Enable dual-mode picolibc builds

keith-packard:
This provides both minsize and release versions of picolibc in the SDK, allowing users to select between them while building their application.


keith-packard:
Looks like this increased the linux x86 toolchain from 1.59GB to 1.66GB?

keith-packard:
GCC has a bug which breaks Zephyr's use case for 'dual' mode. Working on a fix.

stephanosio:
@keith-packard Is this still relevant?

stephanosio:
- Closing as stale. Please re-open if this is still relevant.

Issue #771, opened 2024-06-21
setup.sh wget invocation doesn't work with wget2 in newer distros: "Unknown option 'show-progress'"


karlp:
The wget --show-progress line here
https://github.com/zephyrproject-rtos/sdk-ng/blob/main/scripts/template_setup_posix#L218
doesn't work with the wget2 installed as wget on, for instance, fedora40.

I'm not sure what the right solution here is :)

using `--progress=bar:force` instead of `--show-progress`  _appears_ to be both forwards and backwards compatible, but I don't have extensive tests, just reading the man pages for both versions.

captcha1:
I get this error on Fedora 40 as well.

Maybe should be a separate bug report, but I also notice it does "wget", then "tar" and removes the intermediate file, whereas it's faster and uses less disk space to do : 

wget -O - "${toolchain_uri}" | tar xJ

stephanosio:
After some investigation, it looks like Fedora made a very bad decision to replace `wget` with `wget2`, which is not intended to be a drop-in replacement and "just not ready."

Some noteworthy comments from the upstream wget developers:

> [It has recently come to our attention that Fedora 40 made the decision to ship GNU Wget2 as GNU Wget. While both the projects are maintained by the same developers and Wget2 is indeed supposed to be the natural successor to Wget, it is just not ready. Upstream was NOT asked for its opinion on this matter. If we had been asked about this, we'd have stated our concerns and said that it's a bad idea.](https://gitlab.com/gnuwget/wget2/-/issues/661)

> [We've never claimed Wget2 is ready to be used as Wget. It is a binary called Wget2 and we intend for it to be used in parallel with Wget, atleast for the time being.](https://gitlab.com/gnuwget/wget2/-/issues/661)

Fedora change proposal for wget2: https://discussion.fedoraproject.org/t/f40-change-proposal-wget2-as-wget/96422/6

At this point, it is up to the Fedora team to revert this change, not the downstream projects that expect wget to be wget -- for a workaround, please refer to [the relevant threads on the Fedora forum](https://forums.fedoraforum.org/showthread.php?332571-F40-wget2-breaks-wget-compatibility&s=82fd82821300be0a09f7971e05804a21&p=1884011#post1884011).

stephanosio:
Also note that RHEL 10 will be shipping with `wget` as "real wget," not wget2; only confirming that this was a mistake.

[...]

PR #836, opened 2024-11-20
scripts: Fix using deprecated/removed wget option

jhedberg:
Recent wget versions don't have the --show-progress option anymore and will generate a "wget: unrecognized option '--show-progress'" error (this happens e.g. with latest Fedora). The correct option to use seems to be --progress=bar, which should do the same thing.

### DIFF
--- a/scripts/template_setup_posix
+++ b/scripts/template_setup_posix
@@ -215,7 +215,7 @@ for toolchain in ${inst_toolchains[@]}; do
   echo "Installing '${toolchain}' toolchain ..."
 
   # Download toolchain archive
-  wget -q --show-progress -N -O "${toolchain_filename}" "${toolchain_uri}"
+  wget -q --progress=bar -N -O "${toolchain_filename}" "${toolchain_uri}"
   if [ $? != 0 ]; then
     rm -f "${toolchain_filename}"
     echo "ERROR: Toolchain download failed"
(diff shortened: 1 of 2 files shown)


jhedberg:
Note: I wasn't able to verify this on Windows, but I'd assume the same option should work there as well.

scripts/template_setup_posix
```diff
@@ -215,7 +215,7 @@ for toolchain in ${inst_toolchains[@]}; do
   echo "Installing '${toolchain}' toolchain ..."
 
   # Download toolchain archive
-  wget -q --show-progress -N -O "${toolchain_filename}" "${toolchain_uri}"
+  wget -q --progress=bar -N -O "${toolchain_filename}" "${toolchain_uri}"
```
stephanosio, line 218: This results in progress bar not showing up at all in the "real wget."

stephanosio (changes requested):
**`wget2` is not `wget`**, and Fedora made a mistake of trying to replace `wget` with `wget2`.

See https://github.com/zephyrproject-rtos/sdk-ng/issues/771#issuecomment-2359913981.

Fedora should fix this, not us.

jhedberg:
Thanks @stephanosio, I wasn't aware of this wget vs wget2 distinction. I completely agree Fedora is at fault, however the current end result is that new Zephyr users who follow our getting started guide end up having a negative experience - I wonder if there's something we could still do, e.g. mention this somewhere in the documentation, or even try to work around it in the setup script?

PR #842, opened 2024-11-26
xtensa: Add MediaTek adsp toolchains

andyross:
Add toolchains for mt8186/88 and mt8196.  Note that the directory layout of the overlay is slightly different from existing overlays, conforming to the files as shipped by recent (RJ-2024.3) versions of the Cadence tooling.  crosstools-ng appears to support both, but this is easier to maintain.


contrib/linux_build_toolchain.sh
stephanosio, line 1: This seems to be an unrelated change.

.github/workflows/ci.yml
```diff
@@ -1582,8 +1590,16 @@ jobs:
               PLATFORM_ARGS+="-p intel_adsp/cavs25 "
               ;;
             xtensa-mtk_mt8195_adsp_zephyr-elf)
-              # xtensa-mtk_mt8195_adsp_zephyr-elf is untested because no
-              # upstream user platform is currently available.
+	      # Not merged yet, see Zephyr #81154
+	      #PLATFORM_ARGS+="-p mt8195/mt8195/adsp"
```
stephanosio, line 1594: YAML needs spaces.

stephanosio (commented):
Once the comments here are addressed, I will create a topic branch for testing this.

Note that the CI workflow changes in this PR are not reflected in the actual run because it runs as `pull_request_target`.

andyross:
Fixed

stephanosio:
`topic-mt8196` branch has been created and a test run has been queued:
https://github.com/zephyrproject-rtos/sdk-ng/actions/runs/12060359421

stephanosio (approved):

PR #844, opened 2024-11-29
Mediatek DSP fixups

andyross:
Fixed needed to get a real working toolchain out of the recently merged overlays.  Not quite sure how I convinced myself that was going to work.  My own builds with upstream ct-ng were doing OK, but it's possible it was confounded by a union of the two directory schemes (or maybe we're a little behind?).


andyross:
Yeah... how about a respin that after running `git add` correctly...

andyross:
And I see we now have a build.  And I can confirm it generates working binaries for all four devices with these two in-flight PRs applied:

https://github.com/zephyrproject-rtos/zephyr/pull/82350
https://github.com/zephyrproject-rtos/hal_xtensa/pull/34

Once those merge we can uncomment the test lines in the Github integration.

andyross:
Quick ping here.  The updated HAL and Zephyr PRs have merged now, so the toolchain generated by this PR has been verified to work to build successfully with all the relevant platforms.

overlays/xtensa_mtk_mt8195_adsp/gdb/gdb/xtensa-config.c
```diff
@@ -113,7 +113,7 @@ const xtensa_mask_t xtensa_mask39 = { 1, xtensa_submask39 };
 
 
 /* Register map.  */
-static xtensa_register_t rmap[] =
+xtensa_register_t rmap[] =
```
keith-packard, line 116: oof. global with such a generic name doesn't seem ideal. I assume this is required for ABI though?

andyross: Yeah, these overlays get dropped on top of the binutils source code (that starts out populated with an old Xtensa "sample_controller" file).  It's... not super clean.  FWIW the Espressif-contributed work on LLVM is capable of generating a single toolchain where all these deltas are captured by a config struct instead.  It appears to build Zephyr last I checked, though I haven't had time to see what would be needed to turn it into a "Lite Xtensa SDK".

(Worth nothing that there is no lldb work there that I've seen, so we'd still need to build binutils and gdb I guess.  Though there might be some savings)

stephanosio: > FWIW the Espressif-contributed work on LLVM is capable of generating a single toolchain where all these deltas are captured by a config struct instead. It appears to build Zephyr last I checked, though I haven't had time to see what would be needed to turn it into a "Lite Xtensa SDK".

Interesting. Since LLVM will be part of Zephyr SDK from 0.18.0, we can potentially explore this option. At least, we can drop the GCC builds for the Xtensa toolchain variants whose users do not explicitly require GCC (well, if LLVM works, why should they?)

stephanosio (approved):

stephanosio:
@andyross It looks like macOS is unhappy with these changes ... https://github.com/zephyrproject-rtos/sdk-ng/actions/runs/12245593957/job/34159828915

```
[EXTRA]    Building cross gdb
Error:     /Volumes/Workspace/build/.build/HOST-x86_64-apple-darwin/xtensa-mtk_mt8195_adsp_zephyr-elf/src/gdb/gdb/xtensa-config.c:499:55: error: aliases are not supported on darwin
```

overlays/xtensa_mtk_mt8195_adsp/gdb/gdb/xtensa-config.c
```diff
@@ -496,4 +496,6 @@ static xtensa_register_t rmap[] =
   XTREG_END
 };
 
+extern xtensa_register_t xtensa_rmap[] __attribute__((alias("rmap")));
```
stephanosio, line 499: ^

PR #850, opened 2025-01-11
test: test Dwarf calling convention for i386

kwd-doodling:
this is only for self-test purpose.


kwd-doodling:
useless now. close it.

PR #853, opened 2025-01-15
cmake: introduce gnu and llvm detection when Zephyr toolchain is unset.

tejlmand:
Zephyr SDK v0.18.0 provides both GNU and LLVM based toolchain.

Users may install one or both toolchains.

If ZEPHYR_TOOLCHAIN_VARIANT is not defined, then we lookup first GNU based toolchain, and if found, use this toolchain. Else lookup LLVM based toolchain and use that one.


tejlmand:
@stephanosio I was taking a look at the CI issue, and I notice it's checking out the wrong Zephyr branch:
```
=== Initialized. Now run "west update" inside /__w/sdk-ng/sdk-ng/zephyrproject.
/__w/sdk-ng/sdk-ng/zephyrproject/zephyr /__w/sdk-ng/sdk-ng/zephyrproject
From https://github.com/zephyrproject-rtos/zephyr
 * branch                    collab-sdk-dev -> FETCH_HEAD
Note: switching to 'FETCH_HEAD'.
```

I think it should be using https://github.com/zephyrproject-rtos/zephyr/commits/collab-sdk-0.18-dev/ instead of `collab-sdk-dev`, right ?

stephanosio:
> I think it should be using https://github.com/zephyrproject-rtos/zephyr/commits/collab-sdk-0.18-dev/ instead of `collab-sdk-dev`, right ?

Yes. The `main` branch is still at 0.17.x. Please re-open this pull request against the `topic-clang` branch.

PR #862, opened 2025-02-14
update to qemu v9.2.1 / openocd v0.12.0

nashif:
attempt to move to latest qemu, keeping old version for a few deprecated targets.

- **qemu: update to v9.2.1**
- **qemu: add old version to support nios/leon2**

also update openocd to v0.12.0


nashif:
going another approach, will come in another PR

PR #863, opened 2025-02-17
picolib: Add stdc++ noexcept variant

### LINKED ISSUE #860: Provide -fno-exception libstdc++ variant for picolibc
Hi,

I wanted to switch a bunch of C++ applications from newlibc nano to picolibc but I noticed it resulted in ~4 kB flash usage growth. I realized it was because `nano.specs` replaces `-lstdc++` flag with `-lstdc++_nano`, while `picolibc.specs` does not. The full `libstdc++` contains a lot of code for handling exceptions and stack unwinding, which is often not needed in embedded applications.

I forced the build system to use `-lstdc++_nano` by adding `set_linker_property(PROPERTY c++_library "-lstdc++_nano")` in CMakeLists, and the flash usage dropped by ~10kB.

Would it make sense to default to `lstdc++_nano` when linking with picolibc or at least support the nano variant to be used by applications that disable C++ exceptions?

adigie:
Compile additional variant of libstdc++ with exceptions disabled.

crosstool-ng PR: zephyrproject-rtos/crosstool-ng/pull/32
Fixes #860

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "crosstool-ng"]
 	path = crosstool-ng
-	url = https://github.com/zephyrproject-rtos/crosstool-ng.git
+	url = https://github.com/adigie/crosstool-ng.git
 [submodule "binutils"]
 	path = binutils
 	url = https://github.com/zephyrproject-rtos/binutils-gdb.git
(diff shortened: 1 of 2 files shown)


.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "crosstool-ng"]
 	path = crosstool-ng
-	url = https://github.com/zephyrproject-rtos/crosstool-ng.git
+	url = https://github.com/adigie/crosstool-ng.git
```
keith-packard, line 3: Could we make this use the same picolibcpp.specs file and just replace -lstdc++ with -lstdc++-noexcept? Do we need different header paths or something?

PR #868, opened 2025-02-20
cmake: llvm: fix --config param

DatGizmo:
This fixes 'error: unknown argument: '--config/home/ncs/toolchains/86896046f0/opt/zephyr-sdk/cmake/zephyr/llvm/clang_compiler_rt.cfg

Signed-off-by: Sebastian Wezel <sebastian.wezel@nordicsemi.no>

### DIFF
--- a/cmake/zephyr/llvm/target.cmake
+++ b/cmake/zephyr/llvm/target.cmake
@@ -37,7 +37,5 @@ if(DEFINED triple)
   unset(triple)
 endif()
 
-list(APPEND TOOLCHAIN_C_FLAGS --config
-	${ZEPHYR_SDK_INSTALL_DIR}/cmake/zephyr/llvm/clang_compiler_rt.cfg)
-list(APPEND TOOLCHAIN_LD_FLAGS --config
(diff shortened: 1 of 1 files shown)


DatGizmo:
During testing of the 0.18.0-alpha2 toolchaing with clang I encounters following build error.
```
error: unknown argument: '--config/home/ncs/toolchains/86896046f0/opt/zephyr-sdk/cmake/zephyr/llvm/clang_compiler_rt.cfg
```

I was able to fix it with this change, but not sure if this is the preferred way of doing it.

tejlmand:
this is interesting.

Would be nice with some more info on this, because `TOOLCHAIN_C_FLAGS` / `TOOLCHAIN_LD_FLAGS` are lists, and therefore each item should have a space between them when expanded.

The fact that this doesn't happen makes me worried because the fix in this PR won't help in a case like:
```
list(APPEND TOOLCHAIN_C_FLAGS --foo-flag)
list(APPEND TOOLCHAIN_C_FLAGS --bar-flag)
```

So if each item in the list is not expanded to individual flags then others issues may happen as well.

PR #870, opened 2025-02-21
ci: add needed package for poky 5

nashif:
Packages needed for new version of poky.


nashif:
trivial change and no need to spend time building everything else, currently blocking PRs moving to new poky.

PR #897, opened 2025-04-30
Move to GCC 14.2 and upgrade all needed tooling [DRAFT]

nashif:
- **config: common: do not strip libs**
- **contrib: create workspace if not already created**
- **gdb: use python 3.12**
- **crosstools: save steps**
- **contrib: call upgradeconfig**
- **Move to gcc 14.2 and related tools**
- **remove nios2 support**
- **remove newlib, leave picolibc as the default**
- **overlays: gdb: fix linker api**
- **overlays: add xtensa for picolibc**


keith-packard:
I've rebased the gcc 12 patches to zephyr-14.2.0 and pushed them as keithp-patches-14.2.0 in the gcc repository. Those should clear up the microblazeel build issues.

keith-packard:
There are 606 patches in the 2024.12 release of the foss-for-synopsys-dwc-arc-processors repo on top of releases/gcc-14.2.0 that they have failed to move upstream. Supporting that in the Zephyr SDK seems like a big ask; is there any reason we don't just require that people use the Synopsys toolchain for these processors until upstream gcc has support for their hardware?

nashif:
> There are 606 patches in the 2024.12 release of the foss-for-synopsys-dwc-arc-processors repo on top of releases/gcc-14.2.0 that they have failed to move upstream. Supporting that in the Zephyr SDK seems like a big ask; is there any reason we don't just require that people use the Synopsys toolchain for these processors until upstream gcc has support for their hardware?

@abrodkin we need to figure out the arc64 support for the next SDK release based on 14.2, your input please.

nashif:
Moving this work to a topic branch: topic-gcc_14_2

PR #903, opened 2025-05-09
Topic gcc 14 2

keith-packard:
Toolchain using GCC 14.2 and only picolibc. Using a PR so we only get linux/x86 toolchains built.


nashif:
@keith-packard  I include all changes into one single PR #908, this should move main to the new toolchain

keith-packard:
I've been trying to review the changes, but it's a bit complicated due to rebasing and my own mangling of the git history along with traveling this week.

PR #908, opened 2025-05-13
Move to GCC 14.2

nashif:
GCC 14.2


keith-packard (commented):
Seems like this might slow down builds without helping much

nashif:
> Seems like saving the intermediate crosstool bits won't be that useful?

yep, was a local attemp to speed things up

nashif:
@keith-packard  green now, should we merge to have a baseline for futher devel?

keith-packard (approved):
This provides a useful baseline, let's get started with the rest of the work.

PR #912, opened 2025-05-17
Enable multilib-space functionality

keith-packard:
The bits were missing from crosstool-ng

This series pulls in some other bits:

 * Build libstdc++ without exception support in multilib -Os variants
 * Move picolibc forward to pick up an ARM patch deleting exception bits in asm code.


keith-packard:
Included in #915

PR #926, opened 2025-05-24
gcc/arc: Fix for 32-bit arc shift without barrel shifter

keith-packard:


nashif:
@abrodkin  FYI

stephanosio:
@keith-packard Assuming this is still relevant, could you please rebase?

PR #940, opened 2025-06-06
Riscv fixes

keith-packard:
Use SPEED_OPTIMIZATIONS by default for riscv (avoids compiler bugs with -Os).
Use -mstrict-align by default (avoids emitting misaligned memory operations).

### DIFF
--- a/cmake/zephyr/Kconfig
+++ b/cmake/zephyr/Kconfig
@@ -23,6 +23,9 @@ config PICOLIBC_DEFAULT
 	help
 	  Zephyr SDK >=0.17.1 always uses Picolibc
 
+# libstdc++ is built without exception support in -Os mode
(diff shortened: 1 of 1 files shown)


cmake/zephyr/Kconfig
```diff
@@ -23,6 +23,9 @@ config PICOLIBC_DEFAULT
 	help
 	  Zephyr SDK >=0.17.1 always uses Picolibc
 
+# libstdc++ is built without exception support in -Os mode
+# gcc 14.3 has bugs compiling with -Os on riscv
```
stephanosio, line 27: Is there an upstream GCC PR for this?

stephanosio (approved):

stephanosio:
@keith-packard This needs to be rebased.

PR #946, opened 2025-06-20
ci: when testing, install python 3.12, required by zephyr now

nashif:
Zephyr now requires python 3.12, so install this version when testing
the SDK.


nashif:
> We also need to update `gdb-py` to link against `libpython3.12`.

yep

PR #947, opened 2025-06-20
[0.17-branch] ci: when testing, install python 3.12, required by zephyr now

nashif:
Zephyr now requires python 3.12, so install this version when testing
the SDK.


kartben (approved):

nashif:
actually this is not good, this is just for windows afaik

nashif:
but first let me try removing the if

PR #954, opened 2025-06-26
gcc: Pull RV64GCV multi-lib patch

stephanosio:
This commit pulls in the GCC patch that adds RV64GCV multi-lib built for both `medany` and `large` code models.


stephanosio:
DNM until https://github.com/zephyrproject-rtos/gcc/pull/53 is merged.

PR #961, opened 2025-07-15
meta-zephyr-sdk: Remove qemu_legacy

stephanosio:
This commit removes the `qemu_legacy` recipe for building `qemu-system-sparc` because the main QEMU build is now able to emulate the Zephyr QEMU board.


stephanosio (commented):
[Verified that `qemu_leon3` tests pass with this change](https://gist.github.com/stephanosio/b6ee2703167a6c24bca44f393a471427).

PR #977, opened 2025-07-22
picolibc: Update to version proposed for Zephyr main

keith-packard:
Fast forward to upstream version containing a fix for a C++ incompatibility where isinf and isnan were incorrectly declared as C functions.

This version also contains another useful change to disable the picolibc malloc implementation when built as a Zephyr module. This should improve error messages with Zephyr projects which use malloc but fail to define CONFIG_COMMON_LIBC_MALLOC.

There are also minor cleanups in a few math functions in how FE_INEXACT exceptions are generated and AArch32 A-profile aeabi memcpy functions.

Zephyr PR for this same version:

	https://github.com/zephyrproject-rtos/zephyr/pull/93529


stephanosio:
@keith-packard I just realised that this PR was opened against `v0.17-branch`.

Is this intentional? The picolibc ref at `v0.17-branch` should stay at Zephyr 4.2 level to ensure that 0.17.3 can be used with Zephyr 4.2.

For the latest picolibc changes going into Zephyr `main`, they should go into sdk-ng `main`, not `v0.17-branch` -- Zephyr `main` will be moving to SDK 1.0.0 once released.

keith-packard:
> @keith-packard I just realised that this PR was opened against `v0.17-branch`.
> 
> Is this intentional? The picolibc ref at `v0.17-branch` should stay at Zephyr 4.2 level to ensure that 0.17.3 can be used with Zephyr 4.2.

I intended for this PR to be applied for SDK version 0.17.3; the C++ compatibility issues affect Zephyr 4.2 as well as Zephyr main. The crt0 path issues don't affect Zephyr 4.2 at all -- the bootstrap modules in that version use newlib. But, having that fixed will allow us to migrate the bootstrap modules to picolibc before we switch SDK versions, making the SDK transition more gradual.

> For the latest picolibc changes going into Zephyr `main`, they should go into sdk-ng `main`, not `v0.17-branch` -- Zephyr `main` will be moving to SDK 1.0.0 once released.

I've added a PR for this same picolibc version on top of sdk-ng main so that we've got the same version in all three places (SDK 0.17, SDK main and Zephyr main).

With this change to SDK 0.17.3, I'm able to switch the bootstrap modules to picolibc, apply a few minor GCC 14.3 compat fixes and then run most of the Zephyr test suite using either SDK 0.17 or SDK 0.18, after I hack SDK 0.18 to work without the /gnu directory so I don't need the toolchain path changes in Zephyr.

PR #985, opened 2025-07-27
picolibc: Proposed updates for picolibc

keith-packard:
This aligns the picolibc for 0.18 with the version proposed for SDK version 0.17.3 as well as for the Zephyr picolibc module.

Most of the Zephyr-relevant changes concern using picolibc as a secondary C library and as a module, so there aren't any significant changes for SDK version 0.18.


stephanosio:
Note: https://github.com/zephyrproject-rtos/zephyr/pull/93529 for Zephyr main syncs picolibc to the same version as this.

Issue #851, opened 2025-01-14
Installing Zephyr-sdk v0.18.0-alpha2 breaks bisectability.


tejlmand:
The current Zephyr-sdk v0.18.0-alpha2 breaks bisectability.

The v0.18.0-alpha2 LLVM requires work in https://github.com/zephyrproject-rtos/zephyr/commits/collab-sdk-0.18-dev/, which in general is acceptable, however during a `git bisect` on zephyr/main, then work in this branch (even when merged to zephyr/main in future) will not be available, and therefore it will not be possible to do a `git bisect` which includes revisions before the work found in `collab-sdk-0.18-dev`.

Also it should be possible for users to work with older Zephyr releases, like v3.7.0 and v4.0.0 and still have Zephyr sdk v0.18.0 installed.

Before releasing Zephyr sdk v0.18.0 it should be ensured that older Zephyr revisions can still build, even if this v0.18.0 is also installed.
Especially to guarantee that a git bisect will still work.

Error seen when Zephyr sdk v0.18.0-alpha2 is installed without changes from `collab-sdk-0.18-dev`.

```
-- ZEPHYR_TOOLCHAIN_VARIANT not set, trying to locate Zephyr SDK
-- Found host-tools: zephyr 0.18.0 (/opt/zephyr-sdk-0.18.0-alpha2)
CMake Error at /projects/github/ncs/zephyr/cmake/modules/FindHostTools.cmake:104 (include):
  include could not find requested file:

    /projects/github/ncs/zephyr/cmake/toolchain/zephyr-gnu/generic.cmake
Call Stack (most recent call first):
  /projects/github/ncs/zephyr/cmake/modules/dts.cmake:9 (find_package)
  /projects/github/ncs/zephyr/cmake/modules/zephyr_default.cmake:133 (include)
  /projects/github/ncs/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:66 (include)
  /projects/github/ncs/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:117 (include_boilerplate)
  CMakeLists.txt:5 (find_package)
```

PR #989, opened 2025-07-31
gnu: Add legacy Zephyr compatibility

keith-packard:
In setup.sh, add option to create symlinks from gnu/ so that paths match previous SDK versions.

In Zephyr-sdkConfig.cmake, set ZEPHYR_TOOLCHAIN_VARIANT to zephyr/gnu only when Zephyr sets ZEPHYR_TOOLCHAIN_VARIANT_COMPILER_SUPPORT before finding the SDK module. This will require a change to Zephyr, which is unfortunate. Ideally, we'd use some existing mechanism to detect when to make this switch.

---

Fixes https://github.com/zephyrproject-rtos/sdk-ng/issues/851

### DIFF
--- a/cmake/zephyr/Kconfig
+++ b/cmake/zephyr/Kconfig
@@ -25,5 +25,5 @@ config PICOLIBC_DEFAULT
 
 # libstdc++ is built without exception support in -Os mode
 choice COMPILER_OPTIMIZATIONS
-	default SPEED_OPTIMIZATIONS if ("$(TOOLCHAIN_VARIANT_COMPILER)" = "gnu") && CPP_EXCEPTIONS
+	default SPEED_OPTIMIZATIONS if ("$(TOOLCHAIN_VARIANT_COMPILER)" != "llvm") && CPP_EXCEPTIONS
 endchoice


scripts/template_setup_posix
```diff
@@ -301,6 +307,14 @@ if [ "${do_cmake_pkg}" = "y" ]; then
   echo
 fi
 
+# Create compatibility links for old Zephyr versions
+if [ "${do_old_zephyr}" = "y" ]; then
+  echo "Creating links for old Zephyr compatibility ..."
+  ln -sr gnu/* . || assert_rc "ERROR: Creating toolchain links" 50
+  ln -sr cmake/zephyr/gnu/* cmake/zephyr || assert_rc "ERROR: Creating cmake links" 50
```
stephanosio, line 314: We also need to symlink `hosttools/*` to `.` for host tools.

stephanosio (commented):
On second thought, "compatibility" is probably not the best choice of word here since SDK 1.0.0 is not meant to be "compatible" with Zephyr <= 4.2.

Maybe, s/compatibility/bisectability/ ?

scripts/template_setup_posix
```diff
@@ -69,6 +69,7 @@ usage()
   echo "     all               Install all GNU toolchains"
   echo "  -l                   Install LLVM toolchain"
   echo "  -h                   Install host tools"
+  echo "  -o                   Old zephyr version compatibility"
```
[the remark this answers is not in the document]
keith-packard: Otherwise, yup, make it clear what this does.

scripts/template_setup_posix
keith-packard: Thanks. done.

keith-packard:
I added a commit which inverts the test for when to use -O2 with libstdc++ to preserve bisectability with the existing Zephyr bits. I also added a random cleanup commit which removes a stale comment in the Kconfig file.

PR #995, opened 2025-08-06
gcc: Disable use of SIMD registers in memcpy/memset/memmove expansions

keith-packard:
Avoid using FPU/SIMD registers in code which doesn't explicitly use floating types or SIMD operations.

GCC PR: https://github.com/zephyrproject-rtos/gcc/pull/60


stephanosio (approved):

stephanosio:
@keith-packard https://github.com/zephyrproject-rtos/gcc/pull/60 is merged. Please update the gcc submodule ref to the merged commit.

stephanosio:
Rebased to resolve conflicts after https://github.com/zephyrproject-rtos/sdk-ng/pull/996 is merged

PR #996, opened 2025-08-07
gcc: risc-v: add cr and cf constraint

masz-nordic:
Needed for asm inline to work properly with compressed instructions.

GCC PR: https://github.com/zephyrproject-rtos/gcc/pull/61


stephanosio:
@masz-nordic https://github.com/zephyrproject-rtos/gcc/pull/61 has been merged. Please update the gcc submodule ref in this PR.

PR #997, opened 2025-08-07
ci: Add twister workflow

stephanosio:
This commit adds the "Twister" workflow that can be used to run Zephyr tests on-demand with a specific Zephyr SDK build.


stephanosio (commented):
Tested in sdk-ng-testing repo

PR #1016, opened 2025-09-15
Replace wget on Windows with built-in curl

paulober:
Since Build `17063` (Win 10) Windows includes a curl distribution. This PR adds `curl` as fallback and option to `wget`.

See https://curl.se/windows/microsoft.html for reference.

macOS and Linux people would also benefit from this addition to the posix script.

@tdewey-rpi @will-v-pi

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1931,9 +1931,202 @@ jobs:
         path: test/twister-out/twister.xml
 
   # Post test result check
+  test-downloader-fallback:
+    name: Test Downloader Fallback (${{ matrix.os }}, ${{ matrix.scenario }})
+    needs: [ setup, build-dist-bundle ]
+    runs-on: ${{ matrix.runner }}
+    
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # Test curl fallback on Windows (remove wget)
+          - os: windows
+            runner: windows-2022
+            scenario: curl-only
+            bundle-host: windows-x86_64
+            bundle-archive: 7z
+            toolchain: arm-zephyr-eabi
+            
+          # Test wget on Windows (explicit /dl wget flag)
+          - os: windows
+            runner: windows-2022
+            scenario: wget-explicit
+            bundle-host: windows-x86_64
+            bundle-archive: 7z
+            toolchain: arm-zephyr-eabi
+            
+          # Test curl explicit on Windows
+          - os: windows
+            runner: windows-2022
+            scenario: curl-explicit
+            bundle-host: windows-x86_64
+            bundle-archive: 7z
+            toolchain: arm-zephyr-eabi
+            
+          # Test curl fallback on macOS (no wget by default)
+          - os: macos
+            runner: macos-13
+            scenario: curl-fallback
+            bundle-host: macos-x86_64
+            bundle-archive: tar.xz
+            toolchain: arm-zephyr-eabi
+            
+          # Test curl fallback on Linux (remove wget)
+          - os: linux
+            runner: ubuntu-24.04
+            scenario: curl-only
+            bundle-host: linux-x86_64
+            bundle-archive: tar.xz
+            toolchain: arm-zephyr-eabi
+    
+    defaults:
+      run:
+        shell: bash
+    
+    steps:
+    - name: Check out source code
+      if: ${{ github.event_name != 'pull_request_target' }}
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        persist-credentials: false
+    
+    - name: Check out source code (pull request)
+      if: ${{ github.event_name == 'pull_request_target' }}
+      uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: 0
+        persist-credentials: false
+    
+    - name: Set up test environment (Windows)
+      if: ${{ matrix.os == 'windows' }}
+      run: |
+        # Install required packages (excluding wget for curl-only test)
+        if [ "${{ matrix.scenario }}" == "curl-only" ]; then
+          choco install cmake 7zip
+          echo "Testing curl-only fallback (wget not installed)"
+        elif [ "${{ matrix.scenario }}" == "wget-explicit" ]; then
+          choco install cmake 7zip wget
+          echo "Testing explicit wget flag"
+        elif [ "${{ matrix.scenario }}" == "curl-explicit" ]; then
+          choco install cmake 7zip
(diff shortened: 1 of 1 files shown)


stephanosio:
> Understood. The thing is if you want to integrate zephyr setup into your tooling you have other more convenient options to download the rest so then wget would only be required to run this script which is really not necessary. So what about keeping `wget` the default and give an option via flag / fallback to use curl?

`7z`, another utility not available by default on Windows, will need to be installed or provided in some way for this script to run; so, I do not really understand the desire to use `curl` on the basis that it is already available on Windows -- the script will not be able to run on vanilla Windows anyway.

The question is: why is it so hard to provide `wget` as with `7z`, which you will need to provide anyway, in your tooling?

paulober:
> `7z`, another utility not available by default on Windows, will need to be installed or provided in some way for this script to run; so, I do not really understand the desire to use `curl` on the basis that it is already available on Windows -- the script will not be able to run on vanilla Windows anyway.
> 
> The question is: why is it so hard to provide `wget` as with `7z`, which you will need to provide anyway, in your tooling?

Thanks for the clarification! I completely agree that `7z` is unavoidable — it’s a required tool, so adding that dependency makes sense.

What I’m trying to highlight is that `wget` doesn’t really fall into the same category. Adding it introduces:

* **Extra complexity**: one more binary to download, host, and maintain. (mostly maintain the download links and check that the third party keeps the wget binary updated with the latest security patches by the wget team)
* **Fragile indirection**: depending on a third-party binary hosted in a third-party repo, downloaded through another third-party package manager (e.g. `winget`) that might not even be installed on the users computer (we often had that issue).
* **Little to no benefit** (in our scenario): the main difference is a progress bar, but in non-interactive/headless setups (which are fairly standard these days: ci, IDEs, toolchain installers, fleet management with remote installs) that’s not visible anyway.

In contrast, both macOS and Windows already ship with `curl` by default. Relying on built-in system utilities reduces external dependencies, improves reliability, and avoids unnecessary setup friction.

So my suggestion isn’t to drop `wget` (I think your point is perfect valid and for people who follow the guide and do it manually can benefit of these advantages), but to add `curl` as a **fallback (and opt-in flag)**. That way, users benefit from system-provided tooling where available, and we don’t increase maintenance overhead or complexity.

(in addition: I think there’s a good reason why both Microsoft and Apple decided to include `curl` out of the box, while `wget` is still something users need to fetch manually.)

paulober:
@stephanosio 

> It does actually increase the maintenance overhead for us ensuring the identical behavior across two different tools and dealing with the potential bug reports related to this.

One way to reduce that overhead would be to *standardize* on the tool that is already preinstalled on two of the three supported operating systems (Windows and macOS) and also available on most Linux distributions. Then the only additional case to handle would be Linux systems where `curl` is not already present.

(Also do you really suggest that the tool that is present on almost every car, game console, phone and other systems will be a big source for issues in this still very simple script?)

> `wget` is referenced throughout the Zephyr documentation, including the Getting Started Guide, and installing/providing it should be almost effortless.

Installing `wget` is still less effortless than using tooling that’s already shipped with the operating system and supported by the OS Maintainers (Microsoft and Apple) compared to a random third part person who linked wget to winget. Leveraging built-in utilities avoids additional downloads and setup steps, and improves the out-of-the-box experience for all users.

> Also note that you will need a lot more than just `wget` to be able to set up Zephyr build environment; so, my suggestion is to just provide `wget` alongside many other tools you will need to provide anyway.

Having multiple required dependencies doesn’t make adding another one cost-free — especially when it duplicates functionality already provided by the system. The fact that Zephyr requires Python, for example, could even argue in favor of using Python’s built-in download capabilities, rather than pulling in yet another external tool.

To me, the key point is that `curl` is a modern, system-provided utility, already present on the majority of supported platforms. Using it (either as the default or as a fallback) would reduce external dependencies, lower setup friction, and improve robustness without sacrificing functionality.

stephanosio:
> One way to reduce that overhead would be to standardize on the tool that is already preinstalled on two of the three supported operating systems (Windows and macOS) and also available on most Linux distributions. Then the only additional case to handle would be Linux systems where curl is not already present.

As previously explained, there is a good reason why `wget` was chosen over `curl`.

> Installing wget is still less effortless than using tooling that’s already shipped with the operating system and supported by the OS Maintainers (Microsoft and Apple) compared to a random third part person who linked wget to winget. Leveraging built-in utilities avoids additional downloads and setup steps, and improves the out-of-the-box experience for all users.

Even if you replace `wget` with `curl` here, you will still need to install, just to list a few, `wget`, you need `git`, `cmake`, `ninja-build`, `device-tree-compiler` via `winget` or some other means, all of which are maintained by not "OS Maintainers."

Replacing `wget `with `curl` here does very little to fix the "problems" you speak of when looking at the bigger picture.

> Having multiple required dependencies doesn’t make adding another one cost-free — especially when it duplicates functionality already provided by the system. The fact that Zephyr requires Python, for example, could even argue in favor of using Python’s built-in download capabilities, rather than pulling in yet another external tool.

```
$ ls -lLh /opt/homebrew/bin/wget
-r-xr-xr-x  1 stephanos  admin   461K Sep 22  2024 /opt/homebrew/bin/wget
```

It might as well be free, don't you think?

> My point is not to remove `wget`, but to allow `curl` (which is available by default on both Windows and macOS) as a fallback/option. From a user perspective this:
> 
> * avoids adding an extra dependency that needs to be downloaded, hosted, and maintained,
> * reduces potential setup friction in headless/CI environments,
> * leverages system-provided tooling that both Microsoft and Apple already ship by default. (even most Linux distros)

To set up a working Zephyr build environment, in addition to `wget`, you need `git`, `cmake`, `ninja-build`, `device-tree-compiler`, `gcc`, `gcc-multilib`, `g++-multilib`, and more. Based on your reasoning, with so many extra dependencies, it must be unfathomable to set up and maintain a working Zephyr build environment.

> And on macOS the recommendation is to install a third party package manager which OS integration capability is questionable at least??

Homebrew is a very popular package manager on macOS and that is why we recommend using it. What else would you suggest then? Provide a 10-page instruction in the Getting Started Guide downloading and installing every component one by one?

`wget` is a well established GNU utility that is fairly popular. It is lightweight and incurs practically no overhead. Installing or providing it should be effortless and should not result in any worse overall user experience, practically speaking.

So instead of imposing this seemingly irrational fear of `wget` on others, just install or provide it.

If there is some other reason why you cannot install/provide `wget`, please enlighten me.

stephanosio:
https://github.com/zephyrproject-rtos/sdk-ng/pull/1033 added `wget` to the Zephyr SDK host tools and updated the setup script to use it on macOS and Windows.

PR #1020, opened 2025-09-20
scripts: llvm: Enable static analyzer

tpambor:
Enable static analyzer in clang and provide `clang-extdef-mapping` for Cross Translation Unit (CTU) analysis.


tpambor:
I tested with the build artifact that the static code analysis with CodeChecker now works as expected.

Issue #777, opened 2024-09-06
Clang/LLVM Toolchain Support


stephanosio:
This enhancement issue describes the Clang/LLVM toolchain support plan in the Zephyr SDK.

## Goals

* Provide LLVM Binary Utilities as part of the SDK
* Provide Clang C/C++ Compiler as part of the SDK
* Support SDK Clang/LLVM toolchain in Zephyr

## Specifications

* LLVM Binary Utilities

    * LLVM Binary Utilities shall be an optional component of the SDK that is available separately from the SDK GNU Binutils.
    * The following LLVM Binary Utilities binaries shall be provided <sup>[1]</sup>:
        * dsymutil
        * lld
        * llvm-ar
        * llvm-cov
        * llvm-cxxfilt
        * llvm-dwarfdump
        * llvm-nm
        * llvm-objcopy
        * llvm-objdump
        * llvm-profdata
        * llvm-ranlib (-> llvm-ar)
        * llvm-readelf (-> llvm-readobj)
        * llvm-readobj
        * llvm-size
        * llvm-strip (-> llvm-objcopy)
        * llvm-symbolizer
        * wasm-ld (-> lld)

* Clang Compiler

    * Clang Compiler shall be an optional component of the SDK that is available separately from the SDK GCC.
    * Clang Compiler shall be configured to C and C++ languages.
    * The following Clang compiler binaries shall be provided:
        * clang
        * clang++ (-> clang)
        * clang-cpp (->clang)

* Pre-built Libraries

    * The following pre-built libraries shall be provided:
        * compiler-rt
        * picolibc
        * libunwind
        * libc++abi
        * libc++
    * The pre-built libraries shall be provided for the following Arm targets <sup>[2]</sup>:
        * armv6m_soft_nofp
        * armv7em_hard_fpv4_sp_d16
        * armv7em_hard_fpv5_d16
        * armv7em_soft_nofp
        * armv7m_soft_fpv4_sp_d16
        * armv7m_soft_nofp
        * armv8.1m.main_hard_fp_nomve
        * armv8.1m.main_hard_fpdp_nomve
        * armv8.1m.main_hard_nofp_mve
        * armv8.1m.main_hard_nofp_nomve
        * armv8.1m.main_hard_fp
        * armv8.1m.main_soft_nofp
    * The pre-built libraries shall be provided for the following RISC-V targets <sup>[3]</sup>:
        * rv32e_zicsr_zifencei/ilp32e
[...]

PR #1039, opened 2025-10-29
Restore current Zephyr compatibility

keith-packard:
This patch restores the abilty to use this SDK without any changes to Zephyr. It moves the GNU toolchains back to the top level, places hosttools at the top level and moves the cmake/zephyr/gnu files back to cmake/zephyr.

The goal is to change the sequencing of the restructuring to ease migration. Instead of requiring synchronous changes between SDK and Zephyr, we'll plan on updating Zephyr with the ability to handle either structure, then modify the SDK structure once Zephyr is ready.


keith-packard:
This should now pass if CI is moved to zephyr/main instead of collab-sdk-dev. Who can fix that?

keith-packard:
I've asked the TSC to weigh in on this but haven't heard back from them.

joakim-tjernlund:
> > oh, but what is this PR about then?
> 
> I think Keith did a good job explaining that in the PR description ...
> 
> > Does Clang/Picolib/Zephyr just work then?
> 
> Experimental LLVM/Clang support is currently available in the SDK 1.0-beta1 release and can be tested using the [`collab-sdk-dev branch`](https://github.com/zephyrproject-rtos/zephyr/tree/collab-sdk-dev). See #777 for more details.

OK, starting to get an understanding now. However the changes need lives in various branches and progress to
merge them into Zephyr proper seems to have stalled. I looking for just clone Zephyr "master" and then have both GNU and CLANG. Is that still far out?

PR #1048, opened 2025-12-07
Enable Objective-C language support in config

rodrigopex:
Adding support for Objective-C in Zephyr. I am testing some local options (with a generic arm-none-eabi) that work well with a custom Objective-C runtime based on embedded systems. I would like to explore that with the proper Zephyr SDK supporting it.


keith-packard:
> This is an experimental support only. I would keep that separate for now until we have something really stable. I made that work very well with `arm-none-eabi-gcc` [here](https://github.com/rodrigopex/objective-z). I am trying to integrate that into Zephyr to ensure consistency.

That sounds great.

> The idea is to develop an Objective-C runtime tailored for embedded systems built on Zephyr. The initial work has already been done. It seems to be a promising path. We can control every aspect of the runtime. With that, we can use object-oriented programming fully compatible with Zephyr macros and the kernel, which we can't do with C++.

So developing a modest set of test cases for Zephyr sounds like something you'd probably end up doing as well then, at least to validate this support. Perfect.

> I am working to improve certain memory aspects of the Objective-C runtime by avoiding dynamic allocations or at least isolating them within specific `sys_heap` areas.

I'm sure c++ has similar issues with `new`, although I haven't investigated. My realtime development environments generally forbid any dynamic memory allocation or recursion to avoid surprises...

PR #1084, opened 2026-03-06
ci: twister: Make Linux image version configurable

stephanosio:
This commit updates the Twister workflow to make the Linux Docker image
version used to run tests configurable when manually triggering the
workflow.

Note that this parameter is only relevant when the host type is Linux.


stephanosio (commented):
Tested in https://github.com/zephyrproject-rtos/sdk-ng/actions/runs/22751717371

PR #1111, opened 2026-03-20
xtensa: enable multilib for dc233c and sample_controller{,32}

dcpleung:
This enables multilib support for building GCC. This is needed to support both CALL0 and Windowed ABIs.

The corresponding GCC changes are needed to support multilib where Windowed ABI is the default.

---

GCC PR: https://github.com/zephyrproject-rtos/gcc/pull/71


dcpleung:
Updated the GCC SHA.

stephanosio (approved):

PR #1142, opened 2026-05-15
configs: enable POSIX threading support (--enable-threads=posix)

danahern:
## Summary

Enable C++ threading primitives (`std::mutex`, `std::thread`,
`std::condition_variable`) by building GCC with `--enable-threads=posix`
instead of the default `--enable-threads=no` for bare-metal targets.

This defines `_GLIBCXX_HAS_GTHREADS` in libstdc++ and uses `gthr-posix.h`
as the default threading backend.

### The problem

The Zephyr SDK is built with `--enable-threads=no`, which makes
`std::mutex`, `std::thread`, and `std::condition_variable` unavailable.
Any C++ code using STL threading primitives fails to compile. This blocks
porting C++ libraries that depend on standard threading (ExecuTorch,
XNNPACK, folly, and any Android/Linux C++ code targeting Zephyr).

### Previous attempts

- PR #735: Added `--enable-threads=c11`, merged Oct 2023
- PR #753: Reverted PR #735 (May 2024) — C11 threads caused link failures
  for apps without POSIX threading enabled
- PR #774: Second attempt with weak stubs, never merged

### This approach

Two changes:

1. **`configs/common.config`**: Add `--enable-threads=posix` to
   `CT_CC_GCC_EXTRA_CONFIG_ARRAY`

2. **`stubs/pthread.h`**: A declarations-only stub header installed into
   picolibc's include directory during the build. It provides the type
   definitions and function declarations that `gthr-posix.h` requires
   (`pthread_mutex_t`, `pthread_cond_t`, `pthread_t`, etc.) but **no
   implementations**. The target OS supplies real functions at link time.

This avoids the link failures from previous attempts because:
- The stub contains only declarations, not code
- No pthread symbols are compiled into libgcc or libstdc++
- Applications using threading need `CONFIG_POSIX_API=y` (or equivalent
  weak stubs) to supply the actual pthread implementations

### Impact

- `std::mutex`, `std::thread`, `std::condition_variable` compile and link
- Applications using threading need `CONFIG_POSIX_API=y` or weak pthread
  stubs providing the declared symbols
- No behavior change for code not using C++ threading
- No additional link-time dependencies for non-threading code

### Companion change (Zephyr)

For targets without `CONFIG_POSIX_API`, weak pthread stubs should be added
to Zephyr's `lib/cpp/` to satisfy the linker. These stubs return `ENOSYS`
for operations requiring real threading and are overridden when
`CONFIG_POSIX_API=y`. This will be submitted as a separate PR to the Zephyr
main repository.

## Test plan

- [x] `aarch64-zephyr-elf` toolchain build passes (45m48s on GitHub Actions)
- [x] `arm-zephyr-eabi` toolchain build passes
- [x] `_GLIBCXX_HAS_GTHREADS` is defined as `1` in the built toolchain
- [x] `pthread.h` is installed in the sysroot
- [x] `gthr-default.h` resolves to `gthr-posix.h`
- [x] `std::mutex` / `std::condition_variable` / `std::thread` compile
- [x] `std::condition_variable` out-of-line symbols present in `libstdc++.a`
- [x] Full link succeeds with weak pthread stubs
- [x] **Backward compat**: Plain C compiles (no impact)
- [x] **Backward compat**: C++ without threading compiles (no impact)
- [x] **Backward compat**: No pthread symbols in non-threading object files
- [x] **Backward compat**: `std::atomic` works without pulling in pthread
- [x] **Backward compat**: Including `<mutex>` header without using it: compiles, 0 pthread refs
- [x] **Backward compat**: Using `std::mutex`: compiles, 5 pthread refs (expected)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -556,6 +556,16 @@ jobs:
         git submodule update --depth=1 --recursive gcc_arc
         git submodule update --depth=1 --recursive picolibc
 
+    - name: Install pthread stub headers for POSIX threading support
+      run: |
+        # gthr-posix.h requires <pthread.h> in the sysroot. Install a
+        # declarations-only stub into picolibc's include directory so it
+        # gets installed to the sysroot before the final GCC build.
+        # Real implementations are provided by the target OS (e.g.
+        # Zephyr's CONFIG_POSIX_API).
+        cp ${GITHUB_WORKSPACE}/stubs/pthread.h \
+           ${GITHUB_WORKSPACE}/picolibc/newlib/libc/include/pthread.h
+
     - name: Build crosstool-ng
       run: |
         # Configure macOS build environment
--- a/configs/common.config
+++ b/configs/common.config
@@ -16,7 +16,7 @@ CT_GDB_CUSTOM_LOCATION="${GITHUB_WORKSPACE}/gdb"
 # GCC
 CT_GCC_SRC_CUSTOM=y
 CT_GCC_CUSTOM_LOCATION="${GITHUB_WORKSPACE}/gcc"
-CT_CC_GCC_EXTRA_CONFIG_ARRAY="--with-gnu-ld --with-gnu-as --enable-initfini-array"
+CT_CC_GCC_EXTRA_CONFIG_ARRAY="--with-gnu-ld --with-gnu-as --enable-initfini-array --enable-threads=posix"
 CT_CC_GCC_CORE_EXTRA_CONFIG_ARRAY="--with-gnu-ld --with-gnu-as --enable-initfini-array"
 CT_CC_LANG_CXX=y
 CT_MULTILIB_SPACE=y
--- a/stubs/pthread.h
+++ b/stubs/pthread.h
@@ -1,0 +1,97 @@
+/*
+ * Minimal pthread.h stub for bare-metal targets.
+ *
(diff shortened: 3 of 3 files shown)


cfriedt:
I think my preference would be to not have C++ threads depend on POSIX. It is really easy to make them depend on C11 threads, which should also not pull in POSIX. Bottom line: we cannot make ISO C / C++ depend on POSIX in Zephyr.

So, I like the general direction, just a gentle-NAK because it needs a slight course correction.

For reference #774 (which you found already) and this:

https://github.com/cfriedt/posix-next/blob/main/zephyr/patches/zephyr/c11-thrd-use-sys-thread-and-k-thread.patch

Technically speaking, C++11 threads could also be implemented independently of C11 threads, using zephyr-native APIs. Given the link above, the C++11 threads implementation is also trivial.

cfriedt (changes requested):
POSIX is optional (C++ threads should not require it)

DRNadler:
Ignore this - I see GCC is built with `--enable-threads=zephyr` - Sorry!

Note `--enable-threads=no` makes C++ exception handling unsafe (consider 2 threads throwing).
See: https://bugs.launchpad.net/gcc-arm-embedded/+bug/1905459
Even if the C++ threading API is not supported g++ should still be built with _some_ thread support to make exceptions safe.
Thanks!
Best Regards, Dave

PR #1145, opened 2026-05-22
gcc: add rv64-imafc multilib support

quic-lingutla:
Add multilib support for rv64-imafc ISA to enable systems with single precision floating points.


quic-lingutla:
@stephanosio  Can you please review the patch and enable CI?

PR #1164, opened 2026-09-11
hosttools: add Espressif QEMU

tmedicci:
## Summary
- Add Espressif QEMU as namespaced `qemu-system-espressif-{xtensa,riscv32}` from a pinned [espressif/qemu](https://github.com/espressif/qemu) `esp-develop` snapshot (`esp32`, `esp32s3`, `esp32c3`, `esp32c6`).
- Does not replace stock SDK `qemu-system-xtensa` / `qemu-system-riscv32`.

This is the host-tool foundation for the Zephyr-side PR (CMake discovery + Twister `espressif_qemu`). If it is convenient, a review (and merge if it looks good) before the next feature freeze would help that PR follow in the same window. No pressure if the timing does not work.

## Validated Test plan
- Linux Host Tools + Dist Bundle on this PR (unlabeled default `linux-x86_64`).
- Twister/hello_world against CI SDK artifacts in a fresh west workspace (see Zephyr PR).
- Windows: local MinGW `sdk-build` image + Wine/Win10 nographic hello_world (all four SoCs). macOS aarch64 not validated here.


tmedicci:
Failing test https://github.com/zephyrproject-rtos/sdk-ng/actions/runs/35123471538/job/104914162987?pr=1164 seems to be a connection error. Could it be restarted?

sylvioalves:
This needs forking `espressif/qemu` to `zephyrproject-rtos/espressif-qemu`. @stephanosio @nashif any chances we can have this?

meta-zephyr-sdk/recipes-hosttools/hosttools/hosttools-tarball.bb
sylvioalves, line 38: the espressif prefix misses the firmware prune above so pc-bios ships whole thing, could be removed, right?

```c
rm -rfv ${SDK_OUTPUT}/${SDKPATH}/sysroots/${SDK_SYS}/usr/espressif/share/qemu/edk2*
rm -rfv ${SDK_OUTPUT}/${SDKPATH}/sysroots/${SDK_SYS}/usr/espressif/share/qemu/skiboot.lid
rm -rfv ${SDK_OUTPUT}/${SDKPATH}/sysroots/${SDK_SYS}/usr/espressif/share/qemu/openbios-*
rm -rfv ${SDK_OUTPUT}/${SDKPATH}/sysroots/${SDK_SYS}/usr/espressif/share/qemu/ppc_rom.bin
rm -rfv ${SDK_OUTPUT}/${SDKPATH}/sysroots/${SDK_SYS}/usr/espressif/share/qemu/slof.bin
rm -rfv ${SDK_OUTPUT}/${SDKPATH}/sysroots/${SDK_SYS}/usr/espressif/share/qemu/hppa-firmware.img
```

scripts/build_qemu_espressif.sh
```diff
@@ -0,0 +1,190 @@
...
+"
+
+# Espressif documents --enable-debug and --enable-sdl for local development
+# builds (https://github.com/espressif/esp-toolchain-docs/tree/main/qemu).
+# Neither is shipped here: debug instrumentation is inappropriate for a
+# released host tool, and the SDK runs headless ('-nographic'). Start from the
+# deterministic baseline used by Espressif's own CI and enable only what the
+# Espressif machines need.
+QEMU_FLAGS=" \
+  --bindir=bin \
```
sylvioalves: windows lands in `qemu-espressif/bin/` while qemu and qemu-arc are flat, and the zephyr pr scans that dir without /bin, please re-check.

scripts/build_qemu_espressif.sh
```diff
@@ -0,0 +1,190 @@
...
+if [ "${BUILD_HOST}" == "windows-x86_64" ]; then
+  # Default ninja 'all' also links libslirp wrap tests; only ship the two
+  # emulators used by Zephyr.
+  make -j qemu-system-xtensa.exe qemu-system-riscv32.exe
+else
+  make -j
+fi
+
+# Install QEMU
+make install
```
sylvioalves, line 142: the two-target make is undone by make install pulling in all, perhaps the test skip meant to hold through install?

```suggestion
if [ "${BUILD_HOST}" == "windows-x86_64" ]; then
  # Default ninja 'all' also links libslirp wrap tests; only ship the two
  # emulators used by Zephyr.
  make -j qemu-system-xtensa.exe qemu-system-riscv32.exe
  meson install --no-rebuild
else
  make -j
  make install
fi
```

meta-zephyr-sdk/recipes-devtools/qemu_espressif/files/0002-configure-disable-python-venv-for-meson.patch
```diff
@@ -0,0 +1,37 @@
...
+--- a/configure
++++ b/configure
+@@ -988,12 +988,12 @@ echo "python determined to be '$python'"
+ echo "python version: $($python --version)"
+ 
+-python="$($python -B "${source_path}/python/scripts/mkvenv.py" create pyvenv)"
++python=python3
+ if test "$?" -ne 0 ; then
+     error_exit "python venv creation failed"
+ fi
```
sylvioalves: `python=python3` drops the interpreter configure just resolved and version-checked, and leaves the $? check unreachable

meta-zephyr-sdk/recipes-devtools/qemu_espressif/files/0001-configure-lookup-meson-from-PATH.patch
```diff
@@ -0,0 +1,33 @@
...
+@@ -1002,10 +1002,5 @@ $mkvenv ensuregroup --dir "${source_path}/python/wheels" \
+      ${source_path}/pythondeps.toml meson || exit 1
+ 
+-# At this point, we expect Meson to be installed and available.
+-# We expect mkvenv or pip to have created pyvenv/bin/meson for us.
+-# We ignore PATH completely here: we want to use the venv's Meson
+-# *exclusively*.
+-
+-meson="$(cd pyvenv/bin; pwd)/meson"
++meson="$(command -v meson)"
```
sylvioalves, line 31: command -v in an assignment swallows the failure, so a missing meson configures clean and breaks much later

scripts/build_qemu_espressif.sh
```diff
@@ -0,0 +1,190 @@
...
+    "
+
+  for l in ${QEMU_WIN_LIBS}; do
+    if [ -f "${l}" ]; then
+      cp -f ${l} ${BUILD_PREFIX}/bin
+    fi
+  done
+
+  # Wrap-built slirp is installed as libslirp-0.dll; keep any extra copies.
+  find . -name 'libslirp*.dll' -exec cp -f {} ${BUILD_PREFIX}/bin \;
```
sylvioalves: find -exec exits 0 with no matches, so a missing libslirp dll ships a broken exe instead of failing the build

scripts/build_qemu_espressif.sh
```diff
@@ -0,0 +1,190 @@
...
+    "
+else
+  echo "ERROR: Invalid build host '${BUILD_HOST}'"
+  exit 1
+fi
+
+if [ "${BUILD_HOST}" == "windows-x86_64" ]; then
+  SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+  SLIRP_PATCH="${SCRIPT_DIR}/../meta-zephyr-sdk/recipes-devtools/qemu_espressif/files/0003-meson-windows-shared-libslirp.patch"
+  if grep -A3 "dependency('slirp'" "${BUILD_SOURCE}/meson.build" | grep -q "static: true"; then
```
sylvioalves: if upstream reformats that slirp block the grep misses and the patch is skipped silently, what if you let the patch itself fail instead?

tmedicci: Addressed!

meta-zephyr-sdk/recipes-devtools/qemu_espressif/files/0001-configure-lookup-meson-from-PATH.patch
tmedicci: addressed!

scripts/build_qemu_espressif.sh
tmedicci: Properly checked for empty results!

meta-zephyr-sdk/recipes-devtools/qemu_espressif/files/0002-configure-disable-python-venv-for-meson.patch
tmedicci: Fixed!

scripts/build_qemu_espressif.sh
tmedicci: It doesn't matter because CMake will end up checking both final destination, but I changed it to make it closer to the other implementations.

meta-zephyr-sdk/recipes-hosttools/hosttools/hosttools-tarball.bb
tmedicci: Thanks! Removed them!

scripts/build_qemu_espressif.sh
tmedicci: > the two-target make is undone by make install pulling in all

It isn't. It builds install-only targets. Double-check locally and on CI. The proposed change actually tries to install `trace/trace-events-all` for Windows and fails.

tmedicci:
Failing CI jobs don't seem to be related to this implementation. They are all related to compilers (not touched by this PR) and are not even related to Espressif's.

sylvioalves:
LGTM

